PR #77, opened 2023-01-17
Update podspecs to use `vX.Y.Z` formatting

rebello95:
This is needed to match the formatting we use for releases.


buildbreaker (approved):
Make sure this is conventional for Swift projects

rebello95:
We decided to stick with the existing version formatting, per examples of other packages

PR #92, opened 2023-01-26
Replace `ClientOption` with struct-based config

rebello95:
This PR replaces the `ClientOption` interface and all its conformers with a new struct-based configuration. This has several benefits:
- Enforces that exactly one protocol implementation (Connect or gRPC-Web) is specified at compile-time. Prior, any number could be specified and would cause invalid requests to be serialized
- Enforces that either both the minimum compression size threshold and compression pool are specified for requests, or neither
- Clearer default values (defaults can be seen simply by looking at `ProtocolClientConfig.init`)
- Removes the need for some types like `IdentityCompressionPool`

The original approach was taken to more clearly mirror Go, but this new implementation is better for Swift.


Libraries/Connect/Implementation/ProtocolClientConfig.swift
```diff
@@ -14,51 +14,69 @@
...
-    public let compressionMinBytes: Int?
-    /// The compression type that should be used (e.g., "gzip").
-    /// Requires a matching `compressionPools` entry.
-    public let compressionName: String?
-    /// Compression pools that provide support for the provided `compressionName`, as well as any
-    /// other compression methods that need to be supported for inbound responses.
-    public let compressionPools: [String: CompressionPool]
-    /// Codec to use for serializing/deserializing requests/responses.
+    /// Protocol to use for requests and streams.
+    public let networkProtocol: NetworkProtocol
```
buildbreaker, line 22: I can make this rename as well

rebello95: Cool. I couldn't use protocol since it's a reserved word in swift when lowercased and it'd make ergonomics worse by having to use code tags

PR #129, opened 2023-05-30
Add @available annotations to ConnectMocks streams

mbernson:
This fixes a compilation error that prevented me from using ConnectMocks in my app. The app targets iOS 15, but ConnectMocks didn't compile because it targets iOS 12, and the `@Published` property wrapper that some of its classes use is only available in iOS 13 and higher.

This PR adds `@available(iOS 13.0, *)` to fix the problem.


rebello95 (approved):
Thank you for the contribution!

PR #150, opened 2023-08-08
Replace buf and connect with connectrpc.com

buildbreaker:
Searched for `connect.build` and `buf.build` to replace appropriate usages to `connectrpc.com`

PR #156, opened 2023-08-24
Update Eliza and connectrpc references

smallsamantha:
- Update Eliza protobuf definition to [match connectrpc module](https://buf.build/connectrpc/eliza/file/main:connectrpc/eliza/v1/eliza.proto)
- Update connect.build -> connectrpc.com


rebello95 (approved):
Thanks! Closing #150 in favor of this

PR #160, opened 2023-09-06
Update SwiftProtobuf to 1.23.0

rebello95:
This also pins the SwiftProtobuf plugins in `buf.gen.yaml` files so that they don't change from under CI when a new update is released upstream.

### DIFF
--- a/Examples/ElizaCocoaPodsApp/Podfile.lock
+++ b/Examples/ElizaCocoaPodsApp/Podfile.lock
@@ -1,7 +1,7 @@
 PODS:
   - Connect-Swift (0.6.0):
-    - SwiftProtobuf (~> 1.22.0)
-  - SwiftProtobuf (1.22.0)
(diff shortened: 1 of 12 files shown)


akshayjshah (approved):

Examples/ElizaCocoaPodsApp/Podfile.lock
```diff
@@ -1,7 +1,7 @@
 PODS:
   - Connect-Swift (0.6.0):
-    - SwiftProtobuf (~> 1.22.0)
-  - SwiftProtobuf (1.22.0)
+    - SwiftProtobuf (~> 1.23.0)
+  - SwiftProtobuf (1.23.0)
```
akshayjshah, line 4: I assume that this indentation is autogenerated and correct...

rebello95: Yep

PR #158, opened 2023-09-05
Update references for conformance

smaye81:
This renames all references and usages to the old crosstest repo so that they begin using the new `connectrpc/conformance` repository in the `connectrpc` organization.

Note this is not intended to merge (or build successfully) until the crosstest repo is officially moved.

PR #163, opened 2023-09-08
Update crosstest reference text

smaye81:
This updates references to the crosstest repo to use conformance instead.

Note that the Docker image used for crosstests still uses an old bufbuild image. This is because the test proto packages have changed in the new repo and switching to use one of the new images would cause all existing code to break. This will be fixed in a follow-up PR.


smaye81:
@rebello95 I closed #158 and am going to split it up into two separate PRs. This one can be landed first, which updates most of the text and references to use `conformance` but does not use the new images and new proto packages. Then, I can look into a follow-up PR as to why the new images won't pass tests.

rebello95 (commented):
`CONTRIBUTING.md` needs to be updated

rebello95 (approved):

PR #166, opened 2023-09-12
Update to latest conformance image

smaye81:
This updates the repo to use the latest conformance image, which uses the new organization on Dockerhub as well as the new proto packages (`connectrpc.conformance.v1` instead of `grpc.testing`).

As a part of this update, this attempts a fix at decoding the `ErrorDetails` value, which is returned as a base64-encoded string. The Connect protocol states that implementations should emit unpadded values and that downstream implementations should accept both padded and unpadded values. This adds padding to this string if needed as part of the `ConnectError` decode logic.


Libraries/Connect/Interfaces/ConnectError.swift
```diff
@@ -73,6 +73,23 @@ public struct ConnectError: Swift.Error, Sendable {
...
+                withPad: "=",
+                startingAt: 0
+            )
+
+            self.init(
+                type: try container.decodeIfPresent(String.self, forKey: .type) ?? "",
+                payload: Data(base64Encoded: padded)
+            )
+        }
+
```
smaye81, line 92: @rebello95 Here is the padding fix I mentioned.

Libraries/Connect/Interfaces/ConnectError.swift
```diff
@@ -73,6 +73,23 @@ public struct ConnectError: Swift.Error, Sendable {
             case payload = "value"
         }
 
+        public init(from decoder: Swift.Decoder) throws {
+            let container = try decoder.container(keyedBy: CodingKeys.self)
+
+            // Read the base64-encoded payload and then pad if needed
+            let encodedPayload = try container.decode(String.self, forKey: .payload)
```
rebello95, line 80: Should this be a `decodeIfPresent`? This is changing behavior to throw if `payload` is omitted

Libraries/Connect/Interfaces/ConnectError.swift
rebello95: Do tests cover receiving both a padded and unpadded error?

Libraries/Connect/Interfaces/ConnectError.swift
```diff
@@ -73,6 +73,23 @@ public struct ConnectError: Swift.Error, Sendable {
...
+            let padded = encodedPayload.padding(
+                toLength: ((encodedPayload.count+3)/4)*4,
```
rebello95, line 82:
```suggestion
                toLength: ((encodedPayload.count + 3) / 4) * 4,
```
Also, this is a bit magic -- can you inline an explanation for this?

Libraries/Connect/Interfaces/ConnectError.swift
smaye81: Do you mean the conformance suite? If so, no, they don't account for that. But probably a good one to add when it's refactored.

Libraries/Connect/Interfaces/ConnectError.swift
smaye81: Definitely should. Fixed.

Libraries/Connect/Interfaces/ConnectError.swift
```diff
@@ -73,6 +73,28 @@ public struct ConnectError: Swift.Error, Sendable {
...
+                // Then, pad the string to that length
+                toLength: ((encodedPayload.count + 3) / 4) * 4,
+                withPad: "=",
+                startingAt: 0
+            )
+
+            self.init(
+                type: try container.decodeIfPresent(String.self, forKey: .type) ?? "",
+                payload: Data(base64Encoded: padded)
+            )
```
rebello95, line 95:
```suggestion
            let paddedPayload = encodedPayload.padding(
                // Calculate the nearest multiple of 4 that is >= the length of encodedPayload,
                // then pad the string to that length.
                toLength: ((encodedPayload.count + 3) / 4) * 4,
                withPad: "=",
                startingAt: 0
            )
            self.init(
                type: try container.decodeIfPresent(String.self, forKey: .type) ?? "",
                payload: Data(base64Encoded: paddedPayload)
            )
```

rebello95 (approved):

rebello95 (approved):

rebello95:
Would be nice to add tests for unpadded `ConnectError`s in a follow-up PR

smaye81:
Agreed. I'll put a PR together for that. Thanks for reviewing!

PR #172, opened 2023-09-14
Update podspecs after `connectrpc` organization move

rebello95:


akshayjshah (approved):

akshayjshah:
Nice catch, didn't realize this stuff was in these files too.

PR #174, opened 2023-09-14
Handle `grpc-status` in headers rather than only trailers

### LINKED ISSUE #168: Conflicting gRPC Response
- [x] Searched the repo's previous issues for similar reports.
- [x] Searched the internet for any similar reports.

**Issue Summary**

I am encountering an unusual issue with the response object for a gRPC message. While the [`statusCode`](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md) indicates `ok` and `result` returns success, the headers provide contradictory information. Specifically, the headers contain a status code of `2`, and the `grpc-message` reports an error.

This conflicting information is making it challenging to determine whether an operation was successful or not.

**Detailed Description**

I've confirmed that this issue originates from the client side by using [Postman](https://www.postman.com) to access the same endpoint, which returns the correct information.

**iOS Response:**

In the iOS response, the following points illustrate the inconsistency:

1. The status code is displayed as `ok`/`0`.
2. The headers show a `statusCode` of `2`/`unknown error`.
3. The `grpc-message` in the headers also reports an error.

CleanShot 2023-09-13 at 15 22 30@2x

**Postman Response:**

Contrastingly, when using Postman, I receive the correct response with no contradictory information.

CleanShot 2023-09-13 at 15 28 31@2x

**Possible Cause**

Through my investigation, I believe I've identified a potential cause of this issue. In the `ProtocolClient`, specifically in the `unary(path:request:headers:completion:)` function, the `onResponse` block checks `response.code != .ok` to throw an error. However, in this case, `response.code` is `0`, while the headers indicate `2`. Consequently, `success` is returned instead of raising a `ConnectError` with `.failure`.

rebello95:
"Trailers-only" responses can actually be just headers, in which case we should read `grpc-status` from them.

This PR:
- Updates the logic for both unary and streaming APIs to properly handle cases where "trailers-only" responses are returned using the headers block
- Updates the conformance test suite to test against grpc-go in addition to connect-go in order to catch this case. (grpc-go returns "trailers-only" responses in headers, and connect-go returns them in trailers)

Here's the new logic for gRPC:
- If non-HTTP 200, report an invalid response
- If no body data, assume "trailers-only"
  - Look for the gRPC status and error details in the headers
  - Look for the gRPC status and error details in the trailers
  - If the status is an error, don't process the body

Resolves #168.

### DIFF
--- a/Libraries/ConnectNIO/Internal/GRPCInterceptor.swift
+++ b/Libraries/ConnectNIO/Internal/GRPCInterceptor.swift
@@ -49,19 +49,16 @@ extension GRPCInterceptor: Connect.Interceptor {
                     return response
                 }
 
+                let (grpcCode, connectError) = grpcResult(
+                    fromHeaders: response.headers, trailers: response.trailers
+                )
                 guard let responseData = response.message, !responseData.isEmpty else {
-                    let grpcCode = response.trailers.grpcStatus()
-                    ?? response.headers.grpcStatus() // Trailers-only can be sent in headers block
-                    ?? response.code
                     return Connect.HTTPResponse(
                         code: grpcCode,
                         headers: response.headers,
                         message: response.message,
                         trailers: response.trailers,
-                        error: response.error ?? ConnectError.fromGRPCTrailers(
-                            response.trailers.isEmpty ? response.headers : response.trailers,
-                            code: grpcCode
-                        ),
+                        error: connectError ?? response.error,
                         tracingInfo: response.tracingInfo
                     )
                 }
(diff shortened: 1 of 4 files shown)


akshayjshah:
This probably works, but it isn't my understanding of the protocol.

As I understand it, when receiving a gRPC or gRPC-Web response, we should _first_ check the headers for `grpc-status`. If it's present, the body and trailers should be ignored, and we should look for `grpc-message` and `grpc-status-details-bin` in the headers. If `grpc-status` is _not_ present in the headers, check the body (and if there's no body, look for status/message/details in the trailers).

rebello95:
@akshayjshah thanks for the review, updated.

akshayjshah (approved):
Before merging, can you update the PR description to reflect the updated logic? That way the commit message ends up correct.

PR #177, opened 2023-09-15
Add test for deserializing unpadded base64

smaye81:
This adds a test for deserializing an `ErrorDetail` value with unpadded base64. By default, Swift pads base64-encoding and it seems the only way to not pad is to manually remove the `=`. This adds an option to the `errorData` utility function to remove padding so both scenarios can be tested.

### DIFF
--- a/Tests/ConnectLibraryTests/ConnectTests/ConnectErrorTests.swift
+++ b/Tests/ConnectLibraryTests/ConnectTests/ConnectErrorTests.swift
@@ -31,6 +31,7 @@ final class ConnectErrorTests: XCTestCase {
         XCTAssertEqual(error.unpackedDetails(), [expectedDetails])
         XCTAssertTrue(error.metadata.isEmpty)
     }
+
     func testDeserializingFullErrorAndUnpackingDetailsWithUnpaddedBase64() throws {
         let expectedDetails = Connectrpc_Conformance_V1_SimpleResponse.with {
             $0.hostname = "foobar"
@@ -44,6 +45,7 @@ final class ConnectErrorTests: XCTestCase {
         XCTAssertEqual(error.unpackedDetails(), [expectedDetails])
         XCTAssertTrue(error.metadata.isEmpty)
     }
+
     func testDeserializingFullErrorAndUnpackingMultipleDetails() throws {
         let expectedDetails1 = Connectrpc_Conformance_V1_SimpleResponse.with { $0.hostname = "foo" }
         let expectedDetails2 = Connectrpc_Conformance_V1_SimpleResponse.with { $0.hostname = "bar" }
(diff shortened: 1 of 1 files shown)


Tests/ConnectLibraryTests/ConnectTests/ConnectErrorTests.swift
```diff
@@ -31,7 +31,19 @@ final class ConnectErrorTests: XCTestCase {
         XCTAssertEqual(error.unpackedDetails(), [expectedDetails])
         XCTAssertTrue(error.metadata.isEmpty)
     }
-
```
rebello95, line 34: There should be newlines between functions

Tests/ConnectLibraryTests/ConnectTests/ConnectErrorTests.swift
```diff
@@ -44,7 +56,6 @@ final class ConnectErrorTests: XCTestCase {
         XCTAssertEqual(error.unpackedDetails(), [expectedDetails1, expectedDetails2])
         XCTAssertTrue(error.metadata.isEmpty)
     }
-
```
rebello95, line 47: Same here

Tests/ConnectLibraryTests/ConnectTests/ConnectErrorTests.swift
```diff
@@ -87,17 +98,23 @@ final class ConnectErrorTests: XCTestCase {
...
 
-    private func errorData(expectedDetails: [ProtobufMessage]) throws -> Data {
+    private func errorData(expectedDetails: [ProtobufMessage], pad: Bool = true) throws -> Data {
         // Example error from https://connectrpc.com/docs/protocol/#error-end-stream
         let dictionary: [String: Any] = [
             "code": "unavailable",
             "message": "overloaded: back off and retry",
             "details": try expectedDetails.map { detail in
-                [
+                var val = try detail.serializedData().base64EncodedString()
```
rebello95, line 107: Swift best practices use full words for variable names ("value" and "errorDetail", for example)

Tests/ConnectLibraryTests/ConnectTests/ConnectErrorTests.swift
```diff
@@ -87,17 +98,23 @@ final class ConnectErrorTests: XCTestCase {
...
+                // If pad is false, remove all the padding added by the base64EncodedString function
+                if !pad {
+                    val = val.replacingOccurrences(of: "=", with: "")
+                }
+                let errDetail = [
```
rebello95, line 112:
```suggestion
                return [
```
Can just return this directly

rebello95 (commented):
Test looks good! Just a few stylistic things

rebello95 (approved):

Tests/ConnectLibraryTests/ConnectTests/ConnectErrorTests.swift
smaye81: Ah gotcha, good to know. Thank you.

PR #182, opened 2023-09-22
Relax `Sendable` requirements for interceptors

rebello95:
Instances of interceptors are not expected to be called concurrently from multiple threads, and the `Sendable` requirements currently imposed on them make it challenging for consumers to store state on the interceptors. This PR relaxes the `Sendable` requirements for interceptors, making it easier for consumers to store and access properties on them.

Relatedly, it looks like there may be a Swift bug which prevents annotating properties with `@MainActor` from properly handling sendability: https://forums.swift.org/t/is-this-sendable-warning-correct/55632/2

This PR builds without warnings on Xcode 15 with strict concurrency rules enabled.


rebello95:
@eseay would you mind taking a look at this? 🙏🏽

PR #193, opened 2023-10-10
Fix NIO client-streaming race & flaky test

rebello95:
The new `client_streaming` conformance test that was added in https://github.com/connectrpc/connect-swift/pull/192 turned out to be flaky (failing a couple times out of 100). This only occurred when using the NIO client, and occurred when:

- Client caller opens a stream
- Caller immediately sends one or more chunks of data before the connection is actually established by NIO
- Caller closes the stream and awaits a response

In this case, the stream would correctly send the queued data after the connection was established, but it would never close and would thus never receive a response, causing the test to time out.

This PR resolves this issue by queueing the close action if the connection has not yet been established (in the same way outbound data is queued).

The fix in this PR was validated locally by running the test several hundreds of times without any failures, and it will fix the flakiness seen on `main` after https://github.com/connectrpc/connect-swift/pull/192 merged.


eseay (approved):

eseay:
Makes sense!

PR #194, opened 2023-10-10
Add `ping_pong` conformance tests

rebello95:
Adds tests that conform to the `ping_pong` conformance test: https://github.com/connectrpc/conformance/tree/main#ping_pong


eseay (approved):
Both implementations are very clear to understand. Additionally cross checked with the documentation, and everything checks out.

I do notice that one CI failure, but it doesn't appear to be related to this change.

rebello95:
Having some difficulty figuring out the CI failure because I can't reproduce it locally at all. May come back to this in a bit

PR #205, opened 2023-10-20
Update CI to macOS 13 + Xcode 15.0

rebello95:
Also manually install `colima` to start up Docker since these runners no longer include it out of the box.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,39 +9,39 @@ permissions:
   contents: read
 jobs:
   build-connect-library:
-    runs-on: macos-12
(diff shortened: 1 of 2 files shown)


akshayjshah (approved):
Is testing on older versions of MacOS and XCode not a concern for most Swift projects?

rebello95:
> Is testing on older versions of MacOS and XCode not a concern for most Swift projects?

Not really; teams usually have to upgrade pretty fast in order to support new SDKs. For example, one can't build to an iOS 17 device without Xcode 15. SwiftLint is currently only running CI on Xcode 15 / macOS 13 as well.

PR #209, opened 2023-10-25
Remove `blanket_disable_command` from SwiftLint

rebello95:
This currently warns with:

```
warning: 'blanket_disable_command' is not a valid rule identifier
```


rebello95:
My SwiftLint was outdated 🙃

PR #212, opened 2023-11-03
Reorganize project files & reduce public interface

rebello95:
Reorganizes files in the project to be more clearly and consistently organized by visibility (whether the types within each file should be internal to the package or public). Also reduces the public interface of the module by moving some types to `internal` or `package`.

Note that `package` is specific to SPM, so the keyword cannot be used in conjunction with CocoaPods. For this reason, a few of the types are guarded with `#if COCOAPODS`. It's also worth noting that `ConnectNIO` cannot be consumed via CocoaPods today because NIO does not support CocoaPods. For this reason, the shared gRPC interfaces between `Connect` and `ConnectNIO` don't need to be exposed to CocoaPods at all. Additionally, [`package` requires Swift 5.9](https://github.com/apple/swift-evolution/blob/main/proposals/0386-package-access-modifier.md), so the Swift package version has been updated as well.

New directory structure:

```
.
├── Examples
│   ├── ElizaCocoaPodsApp
│   ├── ElizaSharedSources
│   │   ├── AppSources
│   │   └── GeneratedSources
│   │       └── connectrpc
│   │           └── eliza
│   │               └── v1
│   └── ElizaSwiftPackageApp
│       └── ElizaSwiftPackageApp
├── Libraries
│   ├── Connect
│   │   ├── Internal
│   │   │   ├── Generated
│   │   │   │   └── grpc
│   │   │   │       └── status
│   │   │   │           └── v1
│   │   │   ├── Interceptors
│   │   │   ├── Locks
│   │   │   ├── Streaming
│   │   │   └── Unary
│   │   ├── PackageInternal
│   │   ├── Public
│   │   │   ├── Implementation
│   │   │   │   ├── Clients
│   │   │   │   ├── Codecs
│   │   │   │   └── Compression
│   │   │   └── Interfaces
│   │   │       ├── Interceptors
│   │   │       └── Streaming
│   │   │           ├── AsyncAwait
│   │   │           └── Callbacks
│   │   └── proto
│   │       └── grpc
│   │           └── status
│   │               └── v1
│   ├── ConnectMocks
│   └── ConnectNIO
│       ├── Internal
│       │   └── Extensions
│       └── Public
├── Plugins
│   ├── ConnectMocksPlugin
│   ├── ConnectPluginUtilities
│   └── ConnectSwiftPlugin
└── Tests
    ├── ConnectLibraryTests
    │   ├── ConnectConformance
    │   ├── ConnectMocksTests
    │   ├── ConnectTests
    │   ├── Generated
    │   │   ├── connectrpc
    │   │   │   └── conformance
    │   │   │       └── v1
    │   │   └── server
    │   │       └── v1
    │   └── TestResources
    └── ConnectPluginUtilitiesTests
```


Libraries/ConnectNIO/Internal/GRPCInterceptor.swift
```diff
@@ -186,3 +187,18 @@ extension GRPCInterceptor: StreamInterceptor {
         }
     }
 }
+
+private final class Locked<T>: @unchecked Sendable {
```
rebello95, line 191: Re-implements `Locked` using NIO's locks so that the `Connect.Locked` can stay internal

eseay (approved):

Package.swift
```diff
@@ -75,6 +77,9 @@ let package = Package(
                 "buf.work.yaml",
                 "proto",
                 "README.md",
+            ],
+            swiftSettings: [
+                .unsafeFlags(["-package-name", packageName])
```
eseay, line 82: Have we confirmed that this is not going to cause issues? I have heard and experienced at some point in the past that using unsafe flags causes issues with downstream consumers of the dependency.

eseay (changes requested):
Actually - quick question before approval.

Package.swift
rebello95: Given the Eliza SwiftPM app builds fine with the `Connect` package I think it's okay? Do you know what else we should check? I'm not familiar with those issues

eseay: @rebello95 could we try archiving the Eliza app as a Release IPA? I think that would tell us.

rebello95: Yep, archiving seemed to work fine:
Screenshot 2023-11-13 at 3 45 27 PM

eseay (approved):

Package.swift
rebello95: Seems you were right 🙃 #217

PR #216, opened 2023-11-14
Update CI release job to use Xcode 15

rebello95:
Without this, the release job fails because the `Package.swift` requires a Swift version higher than the GitHub runner's default.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,9 @@ jobs:
     runs-on: macos-13
     steps:
       - uses: actions/checkout@v4
+      - name: Select Xcode version
+        # https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#xcode
+        run: sudo xcode-select --switch /Applications/Xcode_15.0.app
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -11,6 +11,9 @@ jobs:
     runs-on: macos-13
     steps:
       - uses: actions/checkout@v4
+      - name: Select Xcode version
+        # https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#xcode
+        run: sudo xcode-select --switch /Applications/Xcode_15.0.app
```
eseay, line 16: @rebello95 Any reason we wouldn't want to use `Xcode_15.0.1`?

eseay (commented):
Quick q

.github/workflows/release.yml
rebello95: We can, but was just updating to match the rest of the CI jobs for now which use 15.0

rebello95: (This one job was using the default on the GitHub runner, whereas the rest specify a version)

PR #228, opened 2023-12-09
Fix CI jobs

rebello95:
Fixes 2 issues with CI on main:

1. `brew` is not properly installing `docker` because some Python dependencies need to be forcibly unlinked/removed before installing it (more info [here](https://github.com/pypa/pipenv/issues/3831)):

```
==> Pouring maven--3.9.6.ventura.bottle.tar.gz
🍺  /usr/local/Cellar/maven/3.9.6: 92 files, 10.4MB
==> Upgrading python@3.12
  3.12.0 -> 3.12.0_1 

==> Pouring python@3.12--3.12.0_1.ventura.bottle.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3-3.12
Target /usr/local/bin/2to3-3.12
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3-3.12'

To force the link and overwrite all conflicting files:
  brew link --overwrite python@3.12

To list all files that would be deleted:
  brew link --overwrite --dry-run python@3.12
```

2. The conformance test repo is pinned to a version too far in the past, so the sha is not included in the default clone depth and thus cannot be found. This PR bumps the ref (see https://github.com/connectrpc/connect-es/pull/894)


rebello95:
Ok, finally passing 😅

smaye81 (approved):

akshayjshah (approved):

PR #227, opened 2023-12-07
Add support for watchOS 6.0.0+

scottybobandy:
Resolves https://github.com/connectrpc/connect-swift/issues/226

This PR allows connect-swift to be compiled for `watchOS`.

#### Minimum deployment target of `watchOS 6.0.0`:
- Chosen because of the use of `AsyncStream`, which requires `watchOS 6.0.0`. 
- Alternatively, an older version can be targeted, and the @available attribute used to gate `AsyncStream`.

#### As outlined in the linked issue, we cannot import `CFNetwork` on `watchOS`:
- `CFNetwork` import removed from `URLSessionHTTPClient`.
- `fromURLSessionCode(:)` was updated to pattern match on error codes explicitly, taking precedent from `fromHTTPStatus(:)`.

#### Contributing checklist:
- [x] Test suite run and passed on all OS targets.

Issue #226, opened 2023-11-23
connect-swift can't compile for watchOS


scottybobandy:
Hey team 👋 

We're working on adopting `connect-swift` and are running into some issues compiling `connect-swift` for our watchOS app.

For context, we are attempting to perform unary requests from a watchOS app using the `URLSessionHTTPClient` and `grpc-web` protocol. However, `connect-swift` imports `CFNetwork` (only [1 explicit import here](https://github.com/connectrpc/connect-swift/blob/b12276a37e7b78f693f9eb3874188978f85caaf4/Libraries/Connect/Public/Implementation/Clients/URLSessionHTTPClient.swift#L15) AFAIK) which is unavailable on watchOS, so we can't actually compile for our watchOS target.

In addition, [Technote TN3135](https://developer.apple.com/documentation/technotes/tn3135-low-level-networking-on-watchos) describes the 3 specific circumstances when low-level networking is allowed on watchOS. watchOS blocks low-level networking outside of these specific circumstances, impacting aspects of `URLSession` including `URLSessionStreamTask` and `URLSessionWebSocketTask`.

With this limitation on low-level networking, it's likely stream requests will fail even if we can compile and run on watchOS. This is OK for our immediate use-case as we only need to execute unary requests, but a consideration for the future.

Is watchOS support, or at a minimum, the ability to compile and execute unary requests, a reasonable prerequisite for a stable v1.0 release?

scottybobandy:
Quick update with some additional checks required to compile for watchOS:
- `@available(watchOS 6, *)` for accessing `AsyncStream` APIs
- `@available(watchOS 9, *)` for runtime support for parameterized protocol types

rebello95:
👋🏽 thank you for flagging this. I'm out of town through the first week of December, but can investigate when I'm back if @eseay doesn't get to it before then. Alternatively, feel free to submit a PR in the meantime 🙂

rebello95:
Hey @scottybobandy - Thank you for contributing https://github.com/connectrpc/connect-swift/pull/227! I also spent some time investigating this, and I opened a PR to introduce CI validation for watchOS, tvOS, and iOS (rather than only for macOS) as a follow-up: https://github.com/connectrpc/connect-swift/pull/231

[...]

PR #231, opened 2023-12-11
Update CI to Xcode 15.1 & run builds for iOS, watchOS, tvOS

rebello95:
- Adds new CI jobs to build the library against macOS, iOS, tvOS, and watchOS rather than only the default environment (macOS)
- Updates `@available` annotations to properly accommodate these additional platforms
- Updates CI to use Xcode 15.1

Originally the intent was to run tests on each platform, but for some reason GitHub runners are refusing to cooperate and jobs are stalling. Instead, we'll continue running `swift test` which runs tests on the host environment (macOS) and will use `xcodebuild` to build for each platform since it allows for specifying a target SDK/environment (see [this question](https://stackoverflow.com/questions/60245159/how-can-i-build-a-swift-package-for-ios-over-command-line) and [these notes](https://www.jessesquires.com/blog/2021/11/03/swift-package-ios-tests/)).

Follow-up to https://github.com/connectrpc/connect-swift/pull/227 and related to https://github.com/connectrpc/connect-swift/issues/226.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,7 @@ testios: conformanceserverrun ## Run iOS tests
 
 .PHONY: testmacos
 testmacos: conformanceserverrun ## Run macOS tests
-	set -o pipefail && xcodebuild -scheme Connect-Package -destination 'platform=macOS,arch=arm64' test | xcbeautify
+	set -o pipefail && xcodebuild -scheme Connect-Package -destination 'platform=macOS' test | xcbeautify
 	$(MAKE) conformanceserverstop
 
 .PHONY: testtvos
--- a/Makefile
+++ b/Makefile
@@ -80,20 +80,20 @@ $(BIN)/license-headers: Makefile
 
 .PHONY: testios
 testios: conformanceserverrun ## Run iOS tests
-	xcodebuild -scheme Connect-Package -destination 'platform=iOS Simulator,name=iPhone 15,OS=17.0.1' test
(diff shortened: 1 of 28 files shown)


Makefile
eseay, line 1: Not sure if I am doing this unnecessarily, but one thing I always add when running this in my own CI jobs is `set -o pipefail` so that encountered errors will properly fail the job.

Also, I `brew install xcbeautify` and pipe the output of the xcodebuild command to xcbeautify to make the log output way more manageable.

So my invocations would look something like this (note that I normally put the `test` earlier in the command - no logical difference):

`set -o pipefail && xcodebuild test -scheme Connect-Package -destination 'platform=iOS Simulator,name=iPhone 15,OS=17.0.1' | xcbeautify`

eseay (commented):
Just one suggestion/idea.

eseay:
Not sure why the CI is failing, but I do know that we've had issues since moving to Xcode 15 in Actions on my own projects. I see that Xcode 15.1 has just been released, and there's a pending issue to get the actions runners updated with it. 🤞

rebello95:
Yeah I'm not really sure why these are failing either - the job seems to just hang during compilation despite it passing in a reasonable time locally. Can you link the issue you're referring to?

rebello95:
Seems imminent: https://github.com/actions/runner-images/issues/8997

rebello95:
Looks like 15.1 didn't fix the issue on the GitHub runners. I updated the PR and its description to instead build for each platform (which does work on GitHub runners) and run tests the same way we do today on macOS. Tests are green now

eseay (approved):

PR #233, opened 2023-12-14
Test building instead of testing with `xcodebuild` on CI

rebello95:
Not intended for merging.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,7 +59,7 @@ jobs:
           git update-index --refresh --add --remove
           git diff-index --quiet HEAD --
   run-connect-tests:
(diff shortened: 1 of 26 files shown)


eseay:
@rebello95 in order to get it to run on an M1 runner, I think you may need to specify `macos-13-xlarge` as the runner type.

rebello95:
@eseay I think those runners might only be available on paid plans per the [docs](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates):

> Note: Included minutes cannot be used for larger runners. These runners will always be charged for, including in public repositories. For more information, see "[About billing for GitHub Actions](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates)."

PR #239, opened 2024-01-20
Combine headers + trailers into `ConnectError.metadata`

rebello95:
As described by @jhump:

> Currently, users may need to check both headers and trailers when looking for metadata in an error response for an operation with a unary reply (unary and client-stream RPCs). The reason they may have to look in two places is because where the metadata shows up isn't straight-forward - it depends on if the protocol being used and, if gRPC or gRPC-Web, whether the server used a trailers-only response or not. In a trailers-only response, all metadata would show up in the same bag of metadata (which are technically HTTP response headers, but may be classified as "trailers" since they include the status and there will be no subsequent HTTP response trailers in the reply).

This PR updates `ConnectError`'s initialization to combine both headers + trailers into its `metadata` so that it's easier for consumers to query for the keys they need. This also matches connect-go and connect-es.

### DIFF
--- a/Libraries/Connect/PackageInternal/ConnectError+GRPC.swift
+++ b/Libraries/Connect/PackageInternal/ConnectError+GRPC.swift
@@ -31,7 +31,10 @@ extension ConnectError {
         // "Trailers-only" responses can be sent in the headers or trailers block.
         // Check for a valid gRPC status in the headers first, then in the trailers.
         guard let grpcCode = headers?.grpcStatus() ?? trailers?.grpcStatus() else {
-            return (.unknown, nil)
+            return (.unknown, ConnectError(
+                code: .unknown, message: "RPC response missing status", exception: nil,
+                details: [], metadata: [:]
+            ))
         }
 
         if grpcCode == .ok {
@@ -39,7 +42,7 @@ extension ConnectError {
         }
 
         // Combine headers + trailers into metadata to make error parsing easier for consumers,
-        // since gRPC can include error information in both headers and trailers.
+        // since gRPC can include error information in either headers or trailers.
         let metadata = (headers ?? [:]).merging(trailers ?? [:]) { $1 }
         return (grpcCode, .init(
             code: grpcCode,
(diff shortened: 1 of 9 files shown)


Libraries/Connect/PackageInternal/ConnectError+GRPC.swift
```diff
@@ -18,24 +18,36 @@ extension ConnectError {
...
-        if code == .ok {
-            return nil
+    /// - returns: A tuple containing the gRPC status code and an optional error.
+    public static func parseGRPCHeaders(
+        _ headers: Headers?, trailers: Trailers?
+    ) -> (grpcCode: Code, error: ConnectError?) {
+        // "Trailers-only" responses can be sent in the headers or trailers block.
+        // Check for a valid gRPC status in the headers first, then in the trailers.
+        guard let grpcCode = headers?.grpcStatus() ?? trailers?.grpcStatus() else {
+            return (.unknown, nil)
```
jhump, line 34: I'm a little confused by this. If the code is not OK, why would we return a nil error? Should this instead synthesize an error -- perhaps with a message like "RPC response missing status"?

Libraries/Connect/Public/Interfaces/ConnectError.swift
```diff
@@ -110,26 +110,34 @@ extension ConnectError: Swift.Decodable {
...
 extension ConnectError {
-    public static func from(code: Code, headers: Headers, source: Data?) -> Self {
-        let headers = headers.reduce(into: Headers(), { headers, current in
-            headers[current.key.lowercased()] = current.value
-        })
+    public static func from(
+        code: Code, headers: Headers?, trailers: Trailers?, source: Data?
+    ) -> Self {
+        // Combine headers + trailers into metadata to make error parsing easier for consumers,
+        // since gRPC can include error information in both headers and trailers.
```
jhump, line 117:
```suggestion
        // since gRPC can include error information in either headers or trailers.
```

jhump (commented):
Mostly LG. One question and one suggestion.

rebello95:
@eseay would you mind taking a look at this as well? 🙏🏽

eseay (approved):
All makes sense!

PR #241, opened 2024-01-22
Remove unnecessary `async` annotation in demo

rebello95:


rebello95:
@eseay this is green now

eseay (approved):

Issue #252, opened 2024-03-27
Can we prepend onto the path to allow for how the server is set up?


oliverfoggin:
I have a service like `example.v1.ExampleService/SayHello`.

Which gives me a full url like... `https://our.server.com/example.v1.ExampleService/SayHello`.

But the actual path in the server has been configured to be like... `https://our.server.com/some/api/path/example.v1.ExampleService/SayHello`.

I'm not sure how to configure this as it doesn't seem to be something I have access to?

If I add the path onto the `host` when creating the config it is truncated by the time the request is made.

Is this something that we're able to do?

Thanks

oliverfoggin:
Looking further into this the url for the url request is created using...

```
URL(string: path, relativeTo: URL(string: host))
```

But the generated code has a path with an initial `/`. This works by replacing the path on the `relativeTo` url with the `path` provided.

If you remove the initial `/` from the `path` then this works by appending the path onto the existing path of the `relativeTo` url.

So, because the generated code looks like this...

```
@available(iOS 13, *)
internal func `getUser`(request: User_V1_GetUserRequest, headers: Connect.Headers = [:]) async -> ResponseMessage<User_V1_GetUserResponse> {
  return await self.client.unary(path: "/user.v1.UserService/GetUser", idempotencyLevel: .unknown, request: request, headers: headers)
}
```

It will take whatever is provided in the `host` like... `https://google.com/a/b/c/` and create a URL like `https://google.com/user.v1.UserService/GetUser`.

If the generated code drops the initial `/` then we'd get a url like `https://google.com/a/b/c/user.v1.UserService/GetUser` which is what I'm hoping to get to with the project I'm working on.

Is this something that is configurable in the current project? Or would this require some change to make this work?

Thanks

rebello95:
Hey @oliverfoggin - thank you for flagging this issue and for providing examples in #161. I'll investigate this and come up with a resolution. I think the right fix will likely be in how URLs are constructed in the `ProtocolClient`

oliverfoggin:
@rebello95 I agree it definitely needs to be re-looked at.

The quickest/smallest fix would be to revert #161 but that would leave the issue mentioned in there.

Possibly changing the config to take a `host` and `path` and then concatenating them together with the proto service path.

Or providing a `baseURL` and concatenating them.

Either way it should not remove any path elements entered by the developer.

Thanks

rebello95:
[...]

Issue #161, opened 2023-09-07
Add leading slash to methodPath (while generating code)


caio-northfleet-neon:
The ProtocolClient provided as input to the generated `<service>.connect.swift` file has a `host` attribute in its initialization, and the underlying code ends up calling:
```
let url = URL(string: path, relativeTo: URL(string: self.config.host))!
```
Issue is that it does not allow for extra intermediary path elements to stand between `host` and `path`. So, for example, exposing the gRPC service under `https://my.company.com` works fine, but under `https://my.company.com/services` does not. When using the solely exposed `host: "https://my.company.com/services"` property to overcome the issue, unfortunately the underlying URL(string, relativeTo: URL) implementation "swallows" that extra `/services` path in the process.
The workaround, on top of the first workaround used to add the extra path to the `host` attribute, is to add an extra trailing slash to everything, so `host: "https://my.company.com/services/"` does work, the extra `/services` path keeps there and calls work just fine.

I noticed the plugin code below does not include a leading slash to the generated `methodPath`, what would pretty much solve the issue. Note that this leading slash is indeed added to the Go generated code, but not for the Swift and Kotlin generated code using Connect plugins.

Go generated code:
```
const (
	MyService_MyMethod_FullMethodName = "/path.to.MySergice/MyMethod"
)
```

Connect plugin code generator not including the leading slash:
https://github.com/bufbuild/connect-swift/blob/61b79789c5a3179c6cd06b62f41a22e91bceea87/Plugins/ConnectSwiftPlugin/ConnectClientGenerator.swift#L191

Options here are:
- In the plugin, add a leading slash to the generated code.
- In ProtocolClient, expose an extra initialization attribute for `extra path`.
- In ProtocolClient, change the `host` initialization attribute name to `baseUrl` and call `let url = URL(baseUrl+path)` and make sure there's a slash between `baseUrl` and `path`.

rebello95:
Thank you for filing such a detailed issue! I think your suggestion of matching connect-go by adding a leading slash makes sense. Updated here: https://github.com/bufbuild/connect-swift/pull/162

rebello95:
Resolved in #162

oliverfoggin:
@rebello95 I believe this PR is based off an incorrect premise.

The problem is described as the following...

> When using the solely exposed host: "https://my.company.com/services" property to overcome the issue, unfortunately the underlying URL(string, relativeTo: URL) implementation "swallows" that extra /services path in the process.

But this...

[...]

PR #247, opened 2024-02-26
Update conformance runner & tests to v1.0.1

rebello95:
Updates the Connect conformance runner and test suite to [v1.0.1](https://github.com/connectrpc/conformance/releases/tag/v1.0.1) and fixes a few failures from new tests that were added. It also opts out of a few edge cases which are not yet fully specced out / required by the Connect protocol but are now enforced by the conformance runner for the sake of consistency. We may implement logic to conform to these in the future.

PR #253, opened 2024-03-28
Allow additional path extensions on configured hosts

rebello95:
This adds support for specifying hosts with additional path extensions, such as `https://connectrpc.com/a/b/c/`. Previously, `/a/b/c` would be dropped.

Resolves https://github.com/connectrpc/connect-swift/issues/252.

Also related to https://github.com/connectrpc/connect-swift/issues/161.

### DIFF
--- a/Libraries/Connect/Public/Implementation/Clients/ProtocolClient.swift
+++ b/Libraries/Connect/Public/Implementation/Clients/ProtocolClient.swift
@@ -50,7 +50,7 @@ extension ProtocolClient: ProtocolClientInterface {
         var headers = headers
         headers[HeaderConstants.contentType] = ["application/\(config.codec.name())"]
         let request = HTTPRequest<Input>(
-            url: URL(string: path, relativeTo: URL(string: config.host))!,
+            url: config.createURL(forPath: path),
             headers: headers,
             message: request,
             method: .post,
@@ -330,7 +330,7 @@ extension ProtocolClient: ProtocolClientInterface {
         var headers = headers
         headers[HeaderConstants.contentType] = ["application/connect+\(codec.name())"]
         let request = HTTPRequest<Void>(
-            url: URL(string: path, relativeTo: URL(string: self.config.host))!,
+            url: config.createURL(forPath: path),
             headers: headers,
             message: (),
             method: .post,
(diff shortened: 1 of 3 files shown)


oliverfoggin (approved):
Great!

Thanks for making these changes so quickly!

oliverfoggin (changes requested):
Looking at this it appears to add an additional `/` if one already exists.

In this code... 

```
let url = URL(string: "https://foo.com/x/y/z/")!.appendingPathComponent("/a/b/c")
```

The resulting URL is...

```
https://foo.com/x/y/z//a/b/c
```

image

Thanks

oliverfoggin:
It looks like `appendingPathComponent("path")` is deprecated in favour of `appending(path: "path")` but this still has the same issue of the doubled `//`. 😅

image

rebello95:
Thanks for taking a look @oliverfoggin. I'm not seeing the issue you're describing, and I had added a test case for hosts with trailing `/` components which passed (I also just updated the tests to explicitly use the `/a/b/c` examples).

Running locally, this is also what I get:
```
print(ProtocolClientConfig(host: "https://connectrpc.com/a/b/c/")
    .createURL(forPath: rpcPath))

> https://connectrpc.com/a/b/c/connectrpc.conformance.v1.ConformanceService/Unary
```
Which is correct, so I'm not sure why the playground is showing something different.

Can you try checking out this branch as your Connect dependency and seeing if it solves your problem?

> It looks like appendingPathComponent("path") is deprecated in favour of appending(path: "path")

It doesn't look like it's been fully deprecated, but seems like there is a newly preferred version:

Screenshot 2024-03-28 at 6 09 09 AM

However, the new version is not available in the target SDK for Connect, and I think it'd be safer to have one implementation rather than `if #available()` and having 2 codepaths since we aren't getting compilation warnings:

Screenshot 2024-03-28 at 6 09 45 AM

oliverfoggin (approved):

rebello95:
CI failures are unrelated, working with @jhump to resolve

akshayjshah (approved):

oliverfoggin:
I'm on Xcode 15.3 and Sonoma 14.4

Thanks for the update with this. Will check it out and continue on with the project. 😄

oliverfoggin:
@rebello95 

Would it be possible to put a release out for this change? I'm pointing at `main` at the moment to get the change.

Thanks

rebello95:
@oliverfoggin yea I think we can create a release once https://github.com/connectrpc/connect-swift/pull/247 merges

oliverfoggin:
Great! Thanks 😃

PR #265, opened 2024-05-23
Add client-side timeout enforcement

rebello95:
Adds support for triggering client-side timeouts if the client is configured to do so. Resolves several previously opted-out conformance tests.

Resolves https://github.com/connectrpc/connect-swift/issues/267.


akshayjshah (approved):

jhump:
Resolves #267.

Libraries/Connect/Internal/Utilities/TimeoutTimer.swift
```diff
@@ -0,0 +1,57 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+final class TimeoutTimer: @unchecked Sendable {
+    private var hasTimedOut = false
+    private var onTimeout: (() -> Void)?
+    private let queue = DispatchQueue(label: "rpc.connect.Timeout")
```
jhump, line 20: nit: the name looks like it's a timeout for establishing a connection. Maybe `connectrpc.Timeout`?

Libraries/Connect/Internal/Utilities/TimeoutTimer.swift
```diff
@@ -0,0 +1,57 @@
...
+            self?.onTimeout?()
+        }
+    }
+
+    deinit {
+        self.cancel()
+    }
+
+    func start(onTimeout: @escaping () -> Void) {
+        let milliseconds = Int(self.timeout * 1_000)
```
jhump, line 45: Does the `TimeInterval` type not have any methods for unit conversion? If not, maybe name it `timeoutSeconds` to clarify the units?

jhump (approved):
Are timeouts implemented elsewhere for streaming operations?

(Comments are just nits; nothing blocking.)

Libraries/Connect/Internal/Utilities/TimeoutTimer.swift
rebello95: `TimeInterval` is always seconds in Swift. From the [documentation](https://developer.apple.com/documentation/foundation/timeinterval):

> A TimeInterval value is always specified in seconds; it yields sub-millisecond precision over a range of 10,000 years.

Libraries/Connect/Internal/Utilities/TimeoutTimer.swift
rebello95: updated

rebello95:
> Are timeouts implemented elsewhere for streaming operations?

@jhump client-side timeouts are only implemented for unary. I didn't see a failing conformance test for streaming - does one exist?

I'm going to merge this PR to get these updates in, but can follow up with additional additions if needed.

jhump:
>  I didn't see a failing conformance test for streaming - does one exist?

Interesting. There are definitely tests for timeouts with streaming, for all of the stream flavors. But I think it is possible that the server (thanks to the deadline being propagated via request header) sees the timeout and ends up returning the "deadline exceeded" code that the client echos back to the test runner. That's probably a fix we need in the reference server -- so it does _not_ enforce any deadlines, to test that the client does.

The reason this might happen for streams and not for unary in the reference server is that the unary implementation in the server just sleeps for some delay and then returns (no opportunity for a deadline-exceeded error to be observed or returned). But the stream calls must interact with the stream itself, and those interactions can result in errors if the timeout has elapsed, which would then be returned to the client.

rebello95:
> But I think it is possible that the server (thanks to the deadline being propagated via request header) sees the timeout and ends up returning the "deadline exceeded" code that the client echos back to the test runner.

Hmm if I remove the code that sets `grpc-timeout` or `connect-timeout-ms` headers, the only failures I see for streaming are `Deadline Propagation/**` tests whereas I think we'd expect some `Timeouts/**` failures?

> That's probably a fix we need in the reference server -- so it does not enforce any deadlines, to test that the client does.

Makes sense to me, and yea sounds like a good change to add to the runner.

jhump:
> Hmm if I remove the code that sets `grpc-timeout` or `connect-timeout-ms` headers, the only failures I see for streaming are `Deadline Propagation/**` tests whereas I think we'd expect some `Timeouts/**` failures?

Yes, there definitely should be other failures. That's a little surprising, and frankly a little scary. All of those [test cases](https://github.com/connectrpc/conformance/blob/main/internal/app/connectconformance/testsuites/data/timeouts.yaml) expect a `deadline_exceeded` error. So if the server isn't getting a header (and establishing a server-side timeout) and the client isn't enforcing it, how can that possibly be happening? Is there some other transport-level configuration that could possibly be timing out for those cases?

rebello95:
> So if the server isn't getting a header (and establishing a server-side timeout) and the client isn't enforcing it, how can that possibly be happening? Is there some other transport-level configuration that could possibly be timing out for those cases?

I did some more digging and it looks like it's URLSession and SwiftNIO that are triggering these timeouts.

- They're configured here: https://github.com/connectrpc/connect-swift/commit/d01eb7cf3a5fcfea900d6a6ba4fafe1698e5631c
- URLSession errors are mapped here: https://github.com/connectrpc/connect-swift/blob/734f9465462a38666c269d80c20006b61281d1a2/Libraries/Connect/Public/Implementation/Clients/URLSessionHTTPClient.swift#L206-L207
- And NIO is handled here: https://github.com/connectrpc/connect-swift/blob/734f9465462a38666c269d80c20006b61281d1a2/Libraries/ConnectNIO/Internal/ConnectStreamChannelHandler.swift#L176

I fixed it and provided a clearer explanation as to what's happening here: https://github.com/connectrpc/connect-swift/pull/272

Thanks for flagging this!

Issue #270, opened 2024-05-28
client reports incorrect error codes when server returns unexpected content-type, content-encoding


jhump:
These details are not all fully described in the spec (at least [not yet](https://github.com/connectrpc/connectrpc.com/issues/169)). But the conformance suite does have expectations, in an effort to bring consistency to implementations even before the spec updates are complete.

It expects the following:
1. If the content-type has the wrong prefix ("application/" for Connect unary, "application/connect+" for Connect streaming, "application/grpc" for gRPC, and "application/grpc-web" for gRPC-Web), then it appears to be a web server that is _not_ an RPC server. In that case, return an error whose code is derived from the HTTP status code. For gRPC and Connect streaming, where the status code is expected to be 200 OK, this will result in a derived error code of "unknown". This should also be done for any response to a Connect unary response with a non-200 HTTP status whose content-type is not "application/json".
2. If the content-type has a correct prefix, but indicates an unexpected message-codec, the client should report an "internal" error, since the server appears to be an RPC server but sent an invalid response, which is interpreted by the client as an internal server error.
3. If the server sends back a compression encoding that is not supported, the client should report an "internal" error. This is interpreted as an internal server error since a correct server will only use compression that the client is known to accept (via request headers indicating the accepted encodings).
4. If no compression is in use (so no content encoding header or an encoding of "identity"), if a streaming protocol includes the compressed bit set for a message, it should report an "internal" error.

The latter two are consistent with the docs for compression handling with gRPC: https://github.com/grpc/grpc/blob/master/doc/compression.md.

The client implementation in this repo does not behave in the way described above for these scenarios.

Issue #269, opened 2024-05-28
cardinality violations should use error code "unimplemented"


jhump:
The gRPC docs for error codes state that both client and server should use the `unimplemented` code for cardinality violations. See table at the bottom of this doc (you can search for “cardinality violation” in the doc): https://grpc.github.io/grpc/core/md_doc_statuscodes.html.

A cardinality violation is when a stream contains an incorrect number of messages. Specifically, when a response stream for a unary or client-stream RPC contains zero messages with an OK status or more than one message.

The client in this repo does not report an `unimplemented` error code to the application in these cases.

Issue #268, opened 2024-05-28
classification of "trailers-only" responses for gRPC and gRPC-Web is not quite right


jhump:
The logic in the client looks for a "grpc-status" key in response headers to decide if the response is a "trailers-only" response. But the gRPC spec classifies this more rigorously as a headers frame that terminates the HTTP/2 stream. Expanding this to HTTP 1.1, it would be a response that has headers but no response body and no trailers. The current mechanism would misinterpret a response as "trailers-only" if it had an erroneous "grpc-status" header key, even if the response had both a response body and subsequent trailers. To be more compliant with the spec, this client should only consider a response to be "trailers-only" when it observes the end of the response and there was no body and were no trailers. Only then should it query for a "grpc-status" key from the headers.

Issue #222, opened 2023-11-15
Road to v1.0 🥇


rebello95:
**This issue is to track our progress towards tagging a stable v1.0 release of Connect-Swift.**

At this time, we've mostly completed the key tasks that we see as prerequisites for a v1.0:
- [x] Support for the Connect, gRPC, and gRPC-Web protocols
- [x] Interceptors which allow for mutating both typed models and raw data, as well as performing async work
- [x] Support for Connect Unary GET requests
- [x] Compliance with the [conformance test suite](https://github.com/connectrpc/conformance)
- [x] Have several companies using Connect-Swift in production at scale
- [x] Gather more feedback from the community on Connect's usability and APIs so that we can keep them backward-compatible once v1.0 is reached

At this time, our goal is to tag a v1.0 around July 1, 2024. This timeline will allow us to scale adoption, gather feedback, and react to any big news coming out of WWDC with the goal of minimizing the risk of needing to introduce breaking changes shortly after tagging a v1.0. We plan to tag release candidates in the months leading up to this date.

**In the meantime, we encourage the community to stay up-to-date with the latest Connect-Swift version and to share feedback through [GitHub issues](https://github.com/connectrpc/connect-swift/issues) so that we can improve the library over the coming months before tagging a stable release.**

We'll to keep this issue updated as we make progress! :rocket:

rebello95:
Tagged our first RC now that these requirements have been satisfied 🎉 
https://github.com/connectrpc/connect-swift/releases/tag/v1.0.0-rc1

rebello95:
Tagged our first RC now that these requirements have been satisfied 🎉
https://github.com/connectrpc/connect-swift/releases/tag/1.0.0-rc1

rebello95:
1.0 released! 🥳
https://github.com/connectrpc/connect-swift/releases/tag/1.0.0

PR #271, opened 2024-05-29
Resolve all outstanding conformance opt-outs

rebello95:
This PR includes updates to fix all outstanding conformance test failures and removes the corresponding opt-outs. All changes are validated by the conformance test suite.

Many of these fixes are similar to https://github.com/connectrpc/connect-kotlin/pull/248 and https://github.com/connectrpc/connect-kotlin/pull/274.

Resolves https://github.com/connectrpc/connect-swift/issues/268.
Resolves https://github.com/connectrpc/connect-swift/issues/269.
Resolves https://github.com/connectrpc/connect-swift/issues/270.

This should also be one of the final changes before v1.0 https://github.com/connectrpc/connect-swift/issues/222.


Libraries/Connect/Internal/Streaming/ClientOnlyStream.swift
```diff
@@ -0,0 +1,80 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftProtobuf
+
+/// Concrete implementation of `ClientOnlyStreamInterface`.
+///
+/// The complexity around configuring callbacks on this type is an artifact of the library
+/// supporting both callbacks and async/await. This is internal to the package, and not public.
+final class ClientOnlyStream<Input: ProtobufMessage, Output: ProtobufMessage>: @unchecked Sendable {
```
rebello95, line 21: I'm not thrilled about the necessary callback configurations in this class, but it's fairly similar to what's present in `BidirectionalAsyncStream`. The underlying reason is that since we support both classic closures/callbacks and async/await, everything is currently backed by closures/callbacks (with async/await layered on top in separate classes that are available on iOS 13+) which need to be instantiated in a specific order to work with URLSession. If we move away from callbacks at some point or invert this structure, we can simplify these

Libraries/Connect/Internal/Streaming/StreamResult+ClientOnlyStream.swift
```diff
@@ -0,0 +1,54 @@
...
+import Foundation
+
+extension StreamResult {
+    /// Applies some validations which are only relevant for client-only streams.
+    ///
+    /// - parameter receivedMessageCount: The number of response messages that have been received so
+    ///                                   far, including the one being validated.
+    ///
+    /// - returns: The validated stream result, which may have been transformed into an error.
+    func validatedForClientStream(receivedMessageCount: Int) -> Self {
```
rebello95, line 24: For any Kotlin-focused code reviewers 😄 Swift defaults to `internal` so any unannotated functions are not exposed from the package

Libraries/Connect/PackageInternal/Envelope.swift
```diff
@@ -85,6 +85,26 @@ public enum Envelope {
         }
     }
 
+    /// Determines whether the specified data contains more than 1 message.
+    ///
+    /// - parameter packedData: The packed data to analyze.
+    ///
+    /// - returns: True if the data contains more than 1 message.
+    public static func containsMultipleMessages(_ packedData: Data) -> Bool {
```
rebello95, line 93: These are public so the `ConnectGRPC` module can see them. Everything in this directory can be switched to using `package` visibility once it's widely supported

Libraries/Connect/Internal/Interceptors/ConnectInterceptor.swift
```diff
@@ -83,28 +83,51 @@ extension ConnectInterceptor: UnaryInterceptor {
...
         {
-            proceed(HTTPResponse(
-                code: response.code,
-                headers: headers,
-                message: message,
-                trailers: trailers,
-                error: response.error,
+            // If content-type looks like it could be an RPC server's response, consider
+            // this an internal error.
+            let code: Code = contentType.hasPrefix("application/") ? .internalError : .unknown
```
jhump, line 92: Is a non-200 HTTP status handled somewhere else? I expected to see it handled just above, or the HTTP status incorporated here (so deriving a code from the HTTP status instead of hard-coding to "unknown").

Libraries/Connect/Internal/Interceptors/GRPCWebInterceptor.swift
```diff
@@ -61,12 +61,38 @@ extension GRPCWebInterceptor: UnaryInterceptor {
                 response.headers,
                 trailers: response.trailers
             )
+            if grpcCode != .ok || connectError != nil {
```
jhump, line 64: What about if `grpcCode == .ok && connectError != nil`? That would fall into the first case. We should probably change `grpcCode` to be "unknown" in that case. WDYT?

Libraries/Connect/Internal/Streaming/ClientOnlyStream.swift
```diff
@@ -0,0 +1,80 @@
...
+    /// Should be called by the protocol client when a result is received.
+    ///
+    /// - parameter result: The new result that was received.
+    func receive(_ result: StreamResult<Output>) {
+        let receivedMessageCount = self.receivedMessageCount.perform { value in
+            if case .message = result {
+                value += 1
+            }
+            return value
+        }
```
jhump, line 57: I'm not really following this block. I _think_ it means that `receive` must be called 2x in order to see the count bump to 2. Is that right? If so, what if the caller only calls `receive` once? It's not clear to me how we are actually validating that there isn't more than one message in the response, other than returning an error if there is when the application asks for another.

Libraries/Connect/PackageInternal/Envelope.swift
jhump: I'm confused by this function. It looks like it would always return true for gRPC-Web and Connect protocols, since those encode the trailers/status as a second (ostensibly final) message in the stream? Or is there some other layer that sees the "end stream" flag in the next envelope and truncates the data that this helper would see?

Libraries/Connect/Public/Interfaces/Streaming/AsyncAwait/ClientOnlyAsyncStreamInterface.swift
```diff
@@ -39,8 +39,9 @@ public protocol ClientOnlyAsyncStreamInterface<Input, Output> {
     /// - returns: An `AsyncStream` that contains all outputs/results from the stream.
     func results() -> AsyncStream<StreamResult<Output>>
 
-    /// Close the stream. No calls to `send()` are valid after calling `close()`.
-    func close()
+    /// Close the stream and await a response.
+    /// No calls to `send()` are valid after calling `closeAndReceive()`.
+    func closeAndReceive()
```
jhump, line 44: In other Connect and gRPC operations, I'm used to seeing this atomically fuse the `close` and `receive` operations, so that this method returns the one and only response message (or an error). So this seems a little confusingly named since the caller still must call `receive` right after. Or am I misunderstanding this?

Libraries/Connect/Internal/Interceptors/ConnectInterceptor.swift
rebello95: The non-200 cases are falling into one of the `else`/`else if` cases below depending on whether the error response is compressed (`Connect Compressed Error and End-Stream/HTTPVersion:2/TLS:false/error/compressed` is one of these cases). Both of these use `response.code`, which is the Connect code from the HTTP status.

Libraries/Connect/Internal/Interceptors/GRPCWebInterceptor.swift
rebello95: Makes sense to me, will update

Libraries/Connect/Internal/Streaming/ClientOnlyStream.swift
rebello95: This is related to the wiring I alluded to in my earlier comments - `receive` is an internal function called by the protocol client to pass a network chunk to this handle. It's not something called by end-users, it's called by internals _to pass that result on to end users_. I'll rename these to try to provide a bit more clarity (and I'd really like to remove some of this in the future if we can centralize on async/await at some point). Let me know if that clears this up

Libraries/Connect/Public/Interfaces/Streaming/AsyncAwait/ClientOnlyAsyncStreamInterface.swift
rebello95: Related to https://github.com/connectrpc/connect-swift/pull/271#discussion_r1619628243, there is no `receive` function on this interface (that's a function on the internal concrete implementation for wiring), and consumers do not manually fetch results from the stream. Consumers observe responses/results through the `results()` async stream, such as by doing `for await result in stream.results() { ... }`

Libraries/Connect/PackageInternal/Envelope.swift
rebello95: Updated

Libraries/Connect/Internal/Interceptors/GRPCWebInterceptor.swift
rebello95: Updated

Libraries/Connect/Public/Interfaces/Streaming/AsyncAwait/ClientOnlyAsyncStreamInterface.swift
rebello95: That said, I'm also open to better naming suggestions

rebello95: Alternatively, I am open to different naming suggestions. Also not opposed to changing this function to be `async`, removing `results()` from `ClientOnlyAsyncStreamInterface`, and returning the result here. @eseay thoughts?

jhump: > Consumers observe responses/results through the `results()` async stream

If the server sends two messages for a client-stream operation, instead of only one, does the async stream send the first message to the application and _then_ send an error on receipt of the second message? If so, I think the appropriate behavior is to only deliver the message after trailers are received, so you know it was an OK status and exactly one message. That way the application only sees a response message for a well-formed response stream.

Maybe that's what already happens. Sorry if I'm making comments/nitpicks without actually having a clear understanding of what this is doing.

rebello95: > If the server sends two messages for a client-stream operation, instead of only one, does the async stream send the first message to the application and then send an error on receipt of the second message?

Yes, that is what happens now.

> If so, I think the appropriate behavior is to only deliver the message after trailers are received, so you know it was an OK status and exactly one message. That way the application only sees a response message for a well-formed response stream.

To verify I'm understanding correctly, you're suggesting the client pass headers but buffer the body until trailers are received, then verify that only one message was received and return only an error (and no messages) if multiple messages are present? If so, is this something the conformance tests can validate?

jhump: > If so, is this something the conformance tests can validate?

Supposedly it is. It should be complaining if it sees a response message in the `ClientCompatResponse` but the test wasn't expecting one. Is it possible your conformance client is ignoring the initial received message when it subsequently encounters an error?

rebello95: Interesting. I added some logging and this is what I see when running the `Connect Unexpected Responses/HTTPVersion:1/TLS:false/client-stream/multiple-responses` test case:

```
Running 1 tests with reference server for server config {HTTP_VERSION_1, PROTOCOL_CONNECT, TLS:false}...

Received: headers(["content-type": ["application/connect+proto"], "content-length": ["65"]])


Converted into: headers(["content-type": ["application/connect+proto"], "content-length": ["65"]])


URLSession received 65 data bytes


Connect interceptor got 5 data bytes


Connect interceptor not passing message because it is empty


Connect interceptor got 5 data bytes


Connect interceptor not passing message because it is empty


Connect interceptor got 55 data bytes


Connect interceptor received end stream with 50 end stream response bytes


Received: complete(code: Connect.Code.ok, error: nil, trailers: nil)


Converted into: complete(code: Connect.Code.internalError, error: Optional(Connect.ConnectError(code: Connect.Code.unimplemented, message: Optional("unary stream has no messages"), exception: nil, details: [], metadata: [:])), trailers: nil)

Total cases: 1
1 passed, 0 failed
```

So I think what's happening is the Connect/gRPC/gRPC-Web interceptors are are stripping empty messages rather than passing them through to the end client ([here, for example](https://github.com/connectrpc/connect-swift/pull/271/files#diff-2cd1fd07672849e8274188b07c522f3994e56955e376b49af94c73ade98c74dbR217)) and the conformance suite is sending empty messages, so you're right that no message is being returned to the conformance runner. In fact, this is actually falling into the `unary stream has no messages` case because none of the empty messages are being passed through, and the test is passing because both happen to require the same status code.

What should the behavior be in this case? Should the client generally be passing empty messages through?

jhump: Yes, an empty message is a valid message as long as the codec accepts an empty bytes input. If the codec does not accept the empty bytes, then you'd report the error to the client like any other error thrown by the decoding (like from a malformed paylad).

rebello95: Thanks for confirming. I pushed 2 commits:

The first one gets these two tests to fail by no longer stripping empty responses:
- `FAILED: Connect Unexpected Responses/HTTPVersion:1/TLS:false/client-stream/multiple-responses`
- `FAILED: gRPC-Web Unexpected Responses/HTTPVersion:1/TLS:false/trailers-in-body/client-stream/multiple-responses`

The second commit applies buffering to wait for client streams to complete before doing validation and then passing the results to the caller.

rebello95:
Thank you for the detailed review @jhump! 🙏🏽

akshayjshah (approved):

PR #272, opened 2024-05-30
Properly trigger client-side timeouts per-stream

rebello95:
Per the discussion in https://github.com/connectrpc/connect-swift/pull/265#issuecomment-2138141935, this PR avoids setting timeouts on the HTTP client instantiated for each conformance test (`URLSessionHTTPClient` or `NIOHTTPClient`), and instead relies on the `ProtocolClient` to trigger timeouts via request headers (server-triggered timeouts) and client-side timers (client-triggered timeouts). This caused some conformance tests for stream timeouts to fail because the HTTP clients were no longer triggering timeouts internally.

This PR updates the `ProtocolClient` to handle timeouts for streams correctly (both setting outbound request headers and starting local timers for unary requests and streams). Without this change, it was possible for consumers to configure a timeout on the HTTP client (i.e., using `URLSessionConfiguration`), but this would only _partially_ work since outbound request headers such as `grpc-timeout` would not be set. After this change, Connect will handle both cases if `timeout` is set on `ProtocolClientConfig`.

This is similar to the logic in connect-kotlin here: https://github.com/connectrpc/connect-kotlin/commit/76b89e7fc716addd6c80bc4977de6bfbd7241959#diff-cc7f51337bad192f2603a53312146bcd36a7c2bc3fc860327dee53cb3e326749R333-R341

### DIFF
--- a/Libraries/Connect/Internal/Interceptors/ConnectInterceptor.swift
+++ b/Libraries/Connect/Internal/Interceptors/ConnectInterceptor.swift
@@ -120,9 +120,9 @@ extension ConnectInterceptor: StreamInterceptor {
             .requestCompression.map { [$0.pool.name()] }
         headers[HeaderConstants.connectStreamingAcceptEncoding] = self.config
             .acceptCompressionPoolNames()
-//        if let timeout = self.config.timeout {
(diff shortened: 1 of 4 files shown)


jhump (approved):
LGTM

akshayjshah (approved):

PR #284, opened 2024-07-31
Add support for Protobuf Editions

rebello95:
This PR adds support for Protobuf Editions to Connect-Swift plugins. There is no change to generated outputs in this PR, but it adds `supportsEditions` and `supportedEditionRange` specifications in the `Generator` class to ensure the Connect plugins can run with Protobuf files that use Editions.

SwiftProtobuf also deprecated some usages of `Google_Protobuf_Compiler_CodeGeneratorResponse`: 

> 'Google_Protobuf_Compiler_CodeGeneratorResponse.init(files:supportedFeatures:)' is deprecated: Please move your plugin to the CodeGenerator interface

As part of these changes, I migrated over to using `CodeGenerator` in order to be able to specify the new Editions-related variables mentioned above.


Plugins/ConnectPluginUtilities/Generator.swift
```diff
@@ -58,30 +79,75 @@ open class Generator {
...
+                    withExtension: self.outputFileExtension,
+                    using: self.options.fileNaming
+                ),
+                contents: self.resetAndPrintFile(for: descriptor)
+            )
+        }
+    }
+
+    public var supportedEditionRange: ClosedRange<SwiftProtobuf.Google_Protobuf_Edition> {
+        return SwiftProtobufPluginLibrary.DescriptorSet.bundledEditionsSupport
```
jhump, line 142: This seems potentially dangerous: a future edition _could_ introduce service or method-specific changes that need to be handled by the plugin. In such a case, if this plugin code (without the necessary updates) were paired with a newer version of the runtime (that _was_ updated to support that edition) then it would report that the edition is supported even though it may generate incorrect code.

While it's a bit of a maintenance nuisance to have to manually bump the max supported edition, I think that's the safest way to implement this.

Plugins/ConnectSwiftPlugin/ConnectClientGenerator.swift
```diff
@@ -17,26 +17,27 @@ import Foundation
...
 final class ConnectClientGenerator: Generator {
-    private let visibility: String
+    private var visibility = ""
 
-    required init(_ descriptor: FileDescriptor, options: GeneratorOptions) {
-        switch options.visibility {
+    override var outputFileExtension: String {
+        ".connect.swift"
+    }
+
```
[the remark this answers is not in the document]
rebello95: > Is the Generator ever initialized and used on its own?

It's not, I added a comment to clarify this as well

> perhaps we could even consider putting an assertion or something in that superclass outputFileExtension implementation and specify that it must be overridden in subclasses?

This sounds good to me -- I'd rather do that than omit it in subclasses so that it's explicit in each generator. Updated

Plugins/ConnectPluginUtilities/Generator.swift
rebello95: Fair point, updated

Plugins/ConnectPluginUtilities/Generator.swift
```diff
@@ -139,7 +140,13 @@ extension Generator: SwiftProtobufPluginLibrary.CodeGenerator {
     }
 
     public var supportedEditionRange: ClosedRange<SwiftProtobuf.Google_Protobuf_Edition> {
-        return SwiftProtobufPluginLibrary.DescriptorSet.bundledEditionsSupport
+        let minEdition = max(
+            DescriptorSet.bundledEditionsSupport.lowerBound, Google_Protobuf_Edition.legacy
+        )
+        let maxEdition = min(
+            DescriptorSet.bundledEditionsSupport.upperBound, Google_Protobuf_Edition.edition2024
```
jhump, line 147: It is sadly confusing in `descriptor.proto`, which shows 2024 as the latest edition. But 2024 isn't actually fully-defined yet. It's an experimental edition that is still in progress.

If you try to compile a file with it (using buf or the latest v27.2 of protoc) you'll get the following:
```
$ echo 'edition = "2024";' > test.proto && protoc test.proto -o /dev/null
test.proto:1:1: Edition 2024 is later than the maximum supported edition 2023
```

(You can try out 2024 in protoc by passing the `--experimental_editions` flag.)

Since it's incomplete (and 2023 changed quite a bit between early experimental releases like v24 and the final version in v27), we probably should hold off and use 2023 here.

rebello95: Ah, got it. This was also failing without the `min` here for the same reason, but that makes sense to not support 2024 until it's done. Updated

jhump (approved):

Issue #310, opened 2024-10-14
Adoption of Swift 6 features


rebello95:
I'm opening an issue to track some low-hanging fruit with regard to Swift 6 features that we should explore in Connect-Swift:

- [ ] Enforcing strict concurrency: I believe all these warnings were fixed with https://github.com/connectrpc/connect-swift/pull/309, but we should switch the toolchain to Swift 6 to enforce this
- [ ] Typed `throws`: Would it make sense to add `throws(ConnectError)` in some places?
- [ ] Utilize `package` visibility: The files in [this directory](https://github.com/connectrpc/connect-swift/tree/6d0f8edc1f7d68908c34fb6b25a49dce45aa55f9/Libraries/Connect/PackageInternal) are commented as internal-only but are `public` so they can be consumed by `ConnectNIO`. We can make these `package` instead once running on the updated toolchain to remove these from the public interface

We are currently blocked on fully adopting Swift 6 by https://github.com/apple/swift-protobuf/issues/1729.

rebello95:
https://github.com/connectrpc/connect-swift/pull/331 will switch over to compiling with Swift 6

PR #315, opened 2024-10-16
Add deprecation warnings for package-internal APIs

rebello95:
This PR adds deprecation warnings to APIs that are considered package-internal but cannot yet be marked as `package` instead of `public` until we adopt Swift 6 (see https://github.com/connectrpc/connect-swift/issues/310). We can't make these APIs _currently_ deprecated since that will introduce deprecation warnings where they're used within the library, but we can at least add _future deprecation_ warnings which will show up in the IDE as shown below:

Screenshot 2024-10-16 at 11 27 45 AM

The goal of doing this is to help enable us to tag a 1.0 before adopting Swift 6 so we can mark those APIs as `package` afterwards and reasonably consider it a non-breaking change.

### DIFF
--- a/Libraries/Connect/PackageInternal/ConnectError+GRPC.swift
+++ b/Libraries/Connect/PackageInternal/ConnectError+GRPC.swift
@@ -25,6 +25,11 @@ extension ConnectError {
     ///                       passed in the headers block for gRPC-Web.
     ///
     /// - returns: A tuple containing the gRPC status code and an optional error.
+    @available(
+        swift,
+        deprecated: 100.0,
+        message: "This is an internal-only API which will be made package-private in Swift 6."
+    )
     public static func parseGRPCHeaders(
         _ headers: Headers?, trailers: Trailers?
     ) -> (grpcCode: Code, error: ConnectError?) {
--- a/Libraries/Connect/PackageInternal/ConnectError+GRPC.swift
+++ b/Libraries/Connect/PackageInternal/ConnectError+GRPC.swift
@@ -30,11 +30,11 @@ extension ConnectError {
         deprecated: 100.0,
         message: "This is an internal-only API which will be made package-private in Swift 6."
     )
-    public static func parseGRPCHeaders(
+    public static func _parseGRPCHeaders(
         _ headers: Headers?, trailers: Trailers?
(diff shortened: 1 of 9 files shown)


Libraries/Connect/PackageInternal/ConnectError+GRPC.swift
```diff
@@ -25,6 +25,11 @@ extension ConnectError {
     ///                       passed in the headers block for gRPC-Web.
     ///
     /// - returns: A tuple containing the gRPC status code and an optional error.
+    @available(
+        swift,
+        deprecated: 100.0,
```
jhump, line 30: This could be misleading. Is it dangerous to use 1.1 instead? If we end up needing a minor release before we're actually ready to use Swift 6 `package` visibility, could we just bump this out to 1.2? Would that cause havoc?

jhump (approved):
Per separate chat, let's also rename these to start with an underscore -- yet another convention to indicate the symbol is private and that users should avoid it. That way we should be comfortable that a future release that makes them inaccessible will not break anyone.

Otherwise, LGTM.

rebello95:
> Per separate chat, let's also rename these to start with an underscore -- yet another convention to indicate the symbol is private and that users should avoid it. That way we should be comfortable that a future release that makes them inaccessible will not break anyone.

Updated to prefix with underscores 👍🏽

Libraries/Connect/PackageInternal/ConnectError+GRPC.swift
rebello95: This refers to the version of _Swift_ that this will be deprecated in - it's just an arbitrarily high number which doesn't show up in the warning message. Unfortunately, we cannot specify which version of the _package_ something will be deprecated in as part of this macro

jhump: Oh. So why not set it to 6?

rebello95: I tested this out, but it triggers warnings if a consuming project upgrades their `Package.swift` to use Swift 6, which we don't want:

Screenshot 2024-10-19 at 8 42 01 AM

PR #316, opened 2024-10-24
Add deprecation annotations for RPCs marked "deprecated=true"

### LINKED ISSUE #305: Generate @available annotations on deprecated RPCs
Essentially a clone of https://github.com/connectrpc/connect-kotlin/issues/333, but for Swift. Examples below largely influenced by that issue in connect-kotlin.

This would be a very helpful enhancement to the library, as it would allow more effective passive communication from services to their dependents about when certain existing functionality is being deprecated.

There is an ongoing discussion in the [apple/swift-protobuf library here](https://github.com/apple/swift-protobuf/issues/151) about adding this functionality to swift-protobuf itself. Given the age of that discussion, I'd like to consider:
- Is this something that the Connect community would like to try to revitalize?
- Do the challenges that prevented the feature from being implemented some years ago still exist?
- Is there merit in adding this functionality to the connect-swift plugin for its generated services in the absence of it being implemented for models and properties in swift-protobuf?

```
syntax="proto3";

service FooService {
  rpc Foo(FooRequest) returns (FooResponse) {
    // Formally marks the service as being deprecated.
    option deprecated = true;
  }
}

message FooRequest {}
message FooResponse {}
```

Would generate output like:

```diff
public protocol FooServiceClientInterface: Sendable {

-    @available(iOS 13, *)
+    @available(iOS, introduced: 13, deprecated: 13)
    func `foo`(request: FooRequest, headers: Connect.Headers) async -> ResponseMessage<FooResponse>
}
```

There are obviously a few more details to this that we have to consider for other function types, as this only demonstrates a proposed change for the async functions, but all adjustments would be fundamentally similar in nature.

eseay:
This enhances the `ConnectClientGenerator` to inspect the `deprecated` option and annotate the generated service functions accordingly.

- For async/await functions, the existing `@available` annotation, which only exists for iOS, is adjusted accordingly to introduce and deprecate the function at v13, since that is when concurrency was first made available.
  - @rebello95 should we consider adjusting both my deprecation annotation and the existing annotation to account for the other supported platforms of the library?
- Since no pattern was established for generated callback functions, I enhanced their generation to include `@available` annotations for all supported platforms. The "introduced" and "deprecated" versions for those platforms were determined by the minimum supported version of the respective platform defined in `Package.swift`.

Adding the following RPC...
```proto
rpc SayDeprecated(SayRequest) returns (SayResponse) {
  option idempotency_level = NO_SIDE_EFFECTS;
  option deprecated = true;
}
```
...to the ElizaService defined in the [examples-go repository](https://github.com/connectrpc/examples-go/blob/main/proto/connectrpc/eliza/v1/eliza.proto), from which the Eliza sample app sources are generated, and regenerating the Swift service (setting `GenerateCallbackMethods=true` to also demonstrate callback behavior) yields the following result:

```swift
@available(iOS, introduced: 12, deprecated: 12, message: "This function has been marked deprecated.")
@available(macOS, introduced: 10.15, deprecated: 10.15, message: "This function has been marked deprecated.")
@available(tvOS, introduced: 13, deprecated: 13, message: "This function has been marked deprecated.")
@available(watchOS, introduced: 6, deprecated: 6, message: "This function has been marked deprecated.")
@discardableResult
func `sayDeprecated`(request: Connectrpc_Eliza_V1_SayRequest, headers: Connect.Headers, completion: @escaping @Sendable (ResponseMessage<Connectrpc_Eliza_V1_SayResponse>) -> Void) -> Connect.Cancelable

@available(iOS, introduced: 13, deprecated: 13, message: "This function has been marked deprecated.")
func `sayDeprecated`(request: Connectrpc_Eliza_V1_SayRequest, headers: Connect.Headers) async -> ResponseMessage<Connectrpc_Eliza_V1_SayResponse>
```

The generated code for the existing `Say` RPC remains unchanged:

```swift
/// Say is a unary RPC. Eliza responds to the prompt with a single sentence.
@discardableResult
func `say`(request: Connectrpc_Eliza_V1_SayRequest, headers: Connect.Headers, completion: @escaping @Sendable (ResponseMessage<Connectrpc_Eliza_V1_SayResponse>) -> Void) -> Connect.Cancelable

/// Say is a unary RPC. Eliza responds to the prompt with a single sentence.
@available(iOS 13, *)
func `say`(request: Connectrpc_Eliza_V1_SayRequest, headers: Connect.Headers) async -> ResponseMessage<Connectrpc_Eliza_V1_SayResponse>
```

The same additions have been made to the `ConnectMockGenerator`.

Closes #305

### DIFF
--- a/Plugins/ConnectMocksPlugin/ConnectMockGenerator.swift
+++ b/Plugins/ConnectMocksPlugin/ConnectMockGenerator.swift
@@ -117,6 +117,9 @@ final class ConnectMockGenerator: Generator {
 
     private func printCallbackMethodMockImplementation(for method: MethodDescriptor) {
         self.printLine()
+        if let availabilityAnnotation = method.callbackAvailabilityAnnotation() {
+            self.printLine(availabilityAnnotation)
+        }
         if !method.serverStreaming && !method.clientStreaming {
             self.printLine("@discardableResult")
         }
@@ -150,7 +153,9 @@ final class ConnectMockGenerator: Generator {
 
     private func printAsyncAwaitMethodMockImplementation(for method: MethodDescriptor) {
         self.printLine()
-
+        if let availabilityAnnotation = method.asyncAwaitAvailabilityAnnotation() {
+            self.printLine(availabilityAnnotation)
+        }
         self.printLine(
             "\(self.typeVisibility) "
             + method.asyncAwaitSignature(
@@ -217,6 +222,30 @@ private extension MethodDescriptor {
             .init(result: .success(.init())) \
             }
             """
+        }
+    }
+
+    func callbackAvailabilityAnnotation() -> String? {
+        if self.options.deprecated {
+            // swiftlint:disable line_length
+            return """
+            @available(iOS, introduced: 12, deprecated: 12, message: "This function has been marked deprecated.")
+            @available(macOS, introduced: 10.15, deprecated: 10.15, message: "This function has been marked deprecated.")
+            @available(tvOS, introduced: 13, deprecated: 13, message: "This function has been marked deprecated.")
+            @available(watchOS, introduced: 6, deprecated: 6, message: "This function has been marked deprecated.")
+            """
+            // swiftlint:enable line_length
+        } else {
+            return nil
+        }
+    }
+
+    func asyncAwaitAvailabilityAnnotation() -> String? {
+        if self.options.deprecated {
+            // swiftlint:disable:next line_length
+            return "@available(iOS, introduced: 13, deprecated: 13, message: \"This function has been marked deprecated.\")"
+        } else {
+            return nil
         }
     }
 }
--- a/Plugins/ConnectMocksPlugin/ConnectMockGenerator.swift
+++ b/Plugins/ConnectMocksPlugin/ConnectMockGenerator.swift
@@ -229,10 +229,10 @@ private extension MethodDescriptor {
         if self.options.deprecated {
             // swiftlint:disable line_length
             return """
-            @available(iOS, introduced: 12, deprecated: 12, message: "This function has been marked deprecated.")
-            @available(macOS, introduced: 10.15, deprecated: 10.15, message: "This function has been marked deprecated.")
-            @available(tvOS, introduced: 13, deprecated: 13, message: "This function has been marked deprecated.")
-            @available(watchOS, introduced: 6, deprecated: 6, message: "This function has been marked deprecated.")
+            @available(iOS, introduced: 12, deprecated: 12, message: "This RPC has been marked as deprecated in its `.proto` file.")
+            @available(macOS, introduced: 10.15, deprecated: 10.15, message: "This RPC has been marked as deprecated in its `.proto` file.")
+            @available(tvOS, introduced: 13, deprecated: 13, message: "This RPC has been marked as deprecated in its `.proto` file.")
+            @available(watchOS, introduced: 6, deprecated: 6, message: "This RPC has been marked as deprecated in its `.proto` file.")
             """
             // swiftlint:enable line_length
         } else {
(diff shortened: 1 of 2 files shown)


Plugins/ConnectMocksPlugin/ConnectMockGenerator.swift
```diff
@@ -219,4 +224,28 @@ private extension MethodDescriptor {
             """
         }
     }
+
+    func callbackAvailabilityAnnotation() -> String? {
+        if self.options.deprecated {
+            // swiftlint:disable line_length
```
eseay, line 230: @rebello95 @jhump Let me know if this is acceptable. I use SwiftLint in my own repos as well, so I know the rules exist for a purpose; however, I may propose disabling this specific rule in this specific context to improve legibility in the editor and to more closely mimic what will actually be output at the time of generation.

Plugins/ConnectMocksPlugin/ConnectMockGenerator.swift
```diff
@@ -219,4 +224,28 @@ private extension MethodDescriptor {
...
+    }
+
+    func asyncAwaitAvailabilityAnnotation() -> String? {
+        if self.options.deprecated {
+            // swiftlint:disable:next line_length
+            return "@available(iOS, introduced: 13, deprecated: 13, message: \"This function has been marked deprecated.\")"
+        } else {
+            return nil
+        }
+    }
```
eseay, line 250: This function signature and behavior differs from its sibling defined in `ConnectClientGenerator` since the entire type is already annotated with an `@available(iOS 13, *)` in the generated mock. Because of that existing annotation, we only need to produce something if in the case that it's more restrictive than the availability rules defined at the higher scope.

eseay (commented):
A couple in-context notes for y'all to look at as well @rebello95 @jhump. Thanks!

Plugins/ConnectMocksPlugin/ConnectMockGenerator.swift
rebello95: This seems fine to me, doing additional gymnastics would be a bit excessive 😉

Plugins/ConnectMocksPlugin/ConnectMockGenerator.swift
```diff
@@ -219,4 +224,28 @@ private extension MethodDescriptor {
             """
         }
     }
+
+    func callbackAvailabilityAnnotation() -> String? {
+        if self.options.deprecated {
+            // swiftlint:disable line_length
+            return """
+            @available(iOS, introduced: 12, deprecated: 12, message: "This function has been marked deprecated.")
```
rebello95, line 232: Thoughts on changing these messages to "This RPC has been marked as deprecated in its `.proto` file."?

rebello95:
> @rebello95 should we consider adjusting both my deprecation annotation and the existing annotation to account for the other supported platforms of the library?

Yes, I think that would be great 👍🏽 thanks for flagging.

Plugins/ConnectSwiftPlugin/ConnectClientGenerator.swift
```diff
@@ -201,6 +207,30 @@ private extension MethodDescriptor {
         }
     }
 
+    func callbackAvailabilityAnnotation() -> String? {
```
rebello95, line 210: I think all these helper functions can be `private`

rebello95 (approved):
Thanks for implementing this!

Plugins/ConnectSwiftPlugin/ConnectClientGenerator.swift
eseay: I left them without explicit access to align with the functions that were already there; however, I think we're good on not needing to specify access level since the enclosing extension `private extension MethodDescriptor` is already marked.

rebello95: Ah didn't realize these were inside a private extension 👍🏽

rebello95 (approved):

PR #324, opened 2024-11-26
Update `ConnectError.unpackedDetails` documentation comment

### LINKED ISSUE #323: `ConnectError.unpackedDetails` Swift 6 Compatibility
I have a helper extension in one up my packages that depends on Connect-Swift that is defined as follows:

```swift
func unpackFirst<M: SwiftProtobuf.Message>(_ messageType: M.Type) -> M? {
    unpackedDetails().first
}
```

...which is used in a manner similar to what's described in the doc comment for `ConnectError.unpackedDetails`:

`let myMessage = error.unpackFirst(SomeMessageType.self)`.

I am working on migrating my package to support Swift 6, and when building with Xcode 16.1 and compiling with Swift 6, I am receiving the following error in my function's implementation

`unpackedDetails().first -> "Type of expression is ambiguous without a type annotation"`.

I have been able to resolve the issue by adjusting my function implementation as follows:

```swift
func unpackFirst<M: SwiftProtobuf.Message>(_ messageType: M.Type) -> M? {
#if swift(>=6.0)
    let details: [M] = unpackedDetails()
    return details.first
#else
    return unpackedDetails().first
#endif
}
```

eseay:
Updates the documentation comment for `ConnectError.unpackedDetails` to include suggested syntax for accessing the first error of a specific type in a manner that is compatible with Swift 6 and Xcode 16.x.

For more information, please refer to #323.

Closes #323

### DIFF
--- a/Libraries/Connect/Public/Interfaces/ConnectError.swift
+++ b/Libraries/Connect/Public/Interfaces/ConnectError.swift
@@ -34,7 +34,20 @@ public struct ConnectError: Swift.Error, Sendable {
     /// Any decoding errors are ignored, and the detail will simply be omitted from the list.
     ///
     /// To access only the first error of a specific type:
-    /// `let unpackedError: MyError? = error.unpackedDetails().first`
+    ///
+    /// ```
+    /// let unpackedErrors: [MyError] = error.unpackedDetails()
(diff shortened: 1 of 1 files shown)


rebello95 (approved):
Thanks!

PR #331, opened 2025-01-06
Use Swift 6 when available & upgrade to Xcode 16.2

rebello95:
After some discussion in https://github.com/apple/swift-protobuf/issues/1729, I noticed that the compilation error when building with Swift 6 is actually an issue with the config in `Tests/UnitTests/ConnectLibraryTests/buf.gen.yaml` which was still running on `1.25.2` instead of `1.28.2` like other parts of the project.

Changes in this PR:

- Updates that directory to run on `1.28.2` which includes https://github.com/apple/swift-protobuf/commit/3bc7630e56dced84859b5a058b7d39caca538760 and resolves the problem described in the aforementioned issue
- Switches CI to use Xcode 16.2 and macOS 15
- Switches our `Package.swift` to use Swift 6 when available via `swiftLanguageVersions: [.version("6"), .v5]`
- Updates the Eliza SPM app to use Swift 6

This is related to https://github.com/connectrpc/connect-swift/issues/310, but we will not be able to actually implement those until we fully switch to using Swift 6 (a breaking change).


jhump (approved):
Nice!

Package.swift
eseay, line 1: I _think_ what we may want to do instead is keep the `swift-tools-version` set at 5.6 and then specify `swiftLanguageVersions: [.version("6"), .v5]` (instead of `swiftLanguageModes` which is new and Swift 6/Xcode 16 only).

What this will do is allow us to compile with Swift 6 when using the package in Xcode 16, but it will continue to let the package compile with Xcode 15.4 as well. ([docs here](https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/swift6mode#Package-manifest)). I think that's what we want, because if we increase the required `swift-tools-version`, I _believe_ that would technically be considered a breaking change, since any dependent packages that are also set to 5.6 would need to increase their tools version to at least the value specified by Connect-Swift.

I could also be wrong here, but just wanted to share that since I've also been going through this exercise of updating some of my packages to compile with Swift 6 while still wanting them to be usable on Xcode 15.4.


Thoughts?

eseay (commented):
Let me know what you think here. As far as `buf.gen.yaml` and `ci.yaml` - that all looks great.

Package.swift
rebello95: Thanks for this writeup @eseay.

I think you're correct based on the docs you linked, particularly the section on:

> If instead you would just like to use Swift 6 language mode when it’s available (while still continuing to support older modes) you can keep a single Package.swift and specify the version in a compatible manner

I pushed an update with your recommendation 🙏🏽

eseay (approved):
Awesome!

Issue #334, opened 2025-01-09
Circular reference between URLSessionHTTPClient and URLSession


junpluse:
Hi,

There is a circular reference between `URLSessionHTTPClient` and `URLSession` it retains.

https://github.com/connectrpc/connect-swift/blob/20c7871765928bd364dec4414adfafb8b4a150db/Libraries/Connect/Public/Implementation/Clients/URLSessionHTTPClient.swift#L36-L43

According to [Apple documentation](https://developer.apple.com/documentation/foundation/urlsession/1411597-init#parameters), `URLSession`'s delegate object is strongly referenced.

> **Important**
> The session object keeps a strong reference to the delegate until your app exits or explicitly invalidates the session. If you do not invalidate the session by calling the [**invalidateAndCancel()**](https://developer.apple.com/documentation/foundation/urlsession/1411538-invalidateandcancel) or [**finishTasksAndInvalidate()**](https://developer.apple.com/documentation/foundation/urlsession/1407428-finishtasksandinvalidate) method, your app leaks memory until it exits.

Since neither of the above methods are called within `URLSessionHTTPClient`, a memory leak occurs.

In fact, you can confirm the memory leak by launching the [ElizaSwiftPackageApp](https://github.com/connectrpc/connect-swift/tree/main/Examples/ElizaSwiftPackageApp) included in this repository and following the steps below.

1. Tap _Connect (Unary)_
2. Tap _End Chat_
3. Repeat 1-2 several times
4. Click [_Debug Memory Graph_](https://developer.apple.com/documentation/xcode/gathering-information-about-memory-use#Inspect-the-debug-memory-graph) in Xcode's debug area

<img width="1342" alt="Screenshot 2025-01-09 at 13 22 03" src="https://github.com/user-attachments/assets/096af889-ffd6-4507-937e-9a2ec81f43fe" />

You can see that only one `ProtocolClient` remains in memory, whereas the `URLSessionHTTPClient` remains as many as the steps 1-2 above were repeated.

junpluse:
Although not directly related to the above, circular references seem to occur for the following stream and callbacks as well.

https://github.com/connectrpc/connect-swift/blob/20c7871765928bd364dec4414adfafb8b4a150db/Libraries/Connect/Public/Implementation/Clients/ProtocolClient.swift#L179-L191

rebello95:
Thanks for flagging this @junpluse! Should be fixed in https://github.com/connectrpc/connect-swift/pull/335.

PR #335, opened 2025-01-09
Fix retain cycle between `URLSessionHTTPClient` and `URLSession.delegate`

rebello95:
See https://github.com/connectrpc/connect-swift/issues/334 for more information on the underlying issue.

This PR creates a private `URLSessionDelegateWrapper` class to act as an intermediary between `URLSessionHTTPClient` and `URLSession.delegate`. From the inline documentation I added:

```
/// This class exists to avoid a retain cycle between `URLSessionHTTPClient` and its underlying
/// `URLSession.delegate`. Since `URLSession` retains its `delegate` strongly, setting the
/// `URLSessionHTTPClient` directly as the `delegate` will cause a retain cycle:
/// https://developer.apple.com/documentation/foundation/urlsession/1411597-init#parameters
///
/// To work around this, `URLSessionDelegateWrapper` maintains a `weak` reference to the
/// `URLSessionHTTPClient` and passes delegate calls through to it, avoiding the retain cycle.
```

The client now also calls `self.session.finishTasksAndInvalidate()` on `deinit`.

This workaround isn't the _cleanest_, but it does solve the issue and I'm not sure there's a better way of doing it. Open to suggestions, though.

I also made the builder function in the example app `static` to avoid capturing `self` in the builder closures.

With these changes, I was able to confirm the cycle demonstrated in the above issue is resolved in the memory debugger:

Screenshot 2025-01-09 at 9 56 03 AM

Note that I avoided changing the `open func` signatures of `URLSessionHTTPClient` since that would be a breaking API change.

Resolves https://github.com/connectrpc/connect-swift/issues/334.


eseay (approved):
Nice workaround. Sorry for the delay on reviewing here.

rebello95:
Thanks for the review @eseay!

PR #337, opened 2025-01-14
Fix leak in internal implementation of a lock

kinnarr:
While debugging a memory issue in one of our applications, we found a large number of leaked lock instances. This PR fixes the small but rapidly accumulating leak in the internal `Lock` implementation.


rebello95 (approved):
Thank you for this fix!

PR #340, opened 2025-02-04
Prevent duplicate resume continuation on unary request

### LINKED ISSUE #333: UnaryAsyncWrapper resume its continuation more than once
Hi,

We’re seeing quite a lot of random cases of the UnaryAsyncWrapper calling its continuation twice.
From our investigation, we could see that often it is happening when the request was actually cancelled and then its an error.
So something like this flow :
- request is triggered
- request is canceled
- request receives an error (either timeout or our own grpc errors)

`UnaryAsyncWrapper.swift Line 52`

```
Fatal error: SWIFT TASK CONTINUATION MISUSE: send () tried to resume its continuation more than once, returning ResponseMessage<Services_Call_Proto_V1_ListVoicesResponse>(code: Connect.Code.deadlineExceeded, headers: [:], result:
Swift.Result<TheMessagingApp.Services_Call_Proto_V1_ListVoicesResponse, Connect.ConnectError>.failure(Connect.ConnectError(code:
Connect.Code.deadlineExceeded, message: Optional ("timed out"), exception: nil, details: [], metadata: [:])), trailers: [:])!
```

### Reproducibility

I was able to reproduce unconsistently, by
- reducing the timeout to 5s
- Have a network request sent and canceled quickly
- playing with the Network Link Conditioner to slow down network requests to make them fail

It is not consistent so I’m guessing there’s a race depending on how fast we’re cancelling the request.

Screenshot 2025-01-08 at 12 51 29

I was trying to implement a fix, but I’m not sure what would be the best way to check that whether or not the cancelable closure has been canceled or not.

Thanks for any help here 🙏

eseay:
Issue #333 reported an issue where the continuation in the `UnaryAsyncWrapper` was getting resumed twice, which results in a fatal error. Additional details are available in the issue, but the gist is -

When the client is configured to use the gRPC protocol (therefore using swift-nio for networking instead of URLSession), and it is configured with a short timeout internal, when the server responds in error at around the same time that the timeout occurs, the callback that resumes the continuation may fire twice.

I was able to replicate this issue by adjusting the Eliza app's `ProtocolClient` to have a timeout of 0.25s and to point Eliza at a different Connect server (my own team's), such that requests to the Eliza RPCs would result in 404 errors.

This change simply introduces a thread-safe `Bool` to track whether the continuation has been resumed, and if the callback attempts to fire twice, it will throw an `assertionFailure` and return instead of calling `resume` again and causing a fatal error.

Closes #333

### DIFF
--- a/Libraries/Connect/Internal/Unary/UnaryAsyncWrapper.swift
+++ b/Libraries/Connect/Internal/Unary/UnaryAsyncWrapper.swift
@@ -21,7 +21,7 @@ import SwiftProtobuf
 /// https://forums.swift.org/t/how-to-use-withtaskcancellationhandler-properly/54341/37
 /// https://stackoverflow.com/q/71898080
 @available(iOS 13, *)
-actor UnaryAsyncWrapper<Output: ProtobufMessage>: Sendable {
+actor UnaryAsyncWrapper<Output: ProtobufMessage> {
     private var cancelable: Cancelable?
     private let sendUnary: PerformClosure
 
@@ -44,7 +44,12 @@ actor UnaryAsyncWrapper<Output: ProtobufMessage>: Sendable {
         await withTaskCancellationHandler {
             await withCheckedContinuation { continuation in
                 guard !Task.isCancelled else {
-                    continuation.resume(returning: .init(code: .canceled, result: .failure(.canceled())))
+                    continuation.resume(
+                        returning: ResponseMessage(
+                            code: .canceled,
+                            result: .failure(.canceled())
+                        )
+                    )
                     return
                 }
 
@@ -58,8 +63,8 @@ actor UnaryAsyncWrapper<Output: ProtobufMessage>: Sendable {
                 }
             }
         } onCancel: {
-            // When `Task.cancel` signals for this function to be canceled, the underlying function will be canceled
-            // as well.
+            // When `Task.cancel` signals for this function to be canceled,
+            // the underlying function will be canceled as well.
             Task(priority: .high) {
                 await self.cancelable?.cancel()
             }
--- a/Libraries/Connect/Internal/Unary/UnaryAsyncWrapper.swift
+++ b/Libraries/Connect/Internal/Unary/UnaryAsyncWrapper.swift
@@ -41,20 +41,28 @@ actor UnaryAsyncWrapper<Output: ProtobufMessage>: Sendable {
     ///
     /// - returns: The response/result of the request.
     func send() async -> ResponseMessage<Output> {
-        return await withTaskCancellationHandler(operation: {
-            return await withCheckedContinuation { continuation in
-                if Task.isCancelled {
-                    continuation.resume(
-                        returning: .init(code: .canceled, result: .failure(.canceled()))
-                    )
-                } else {
-                    self.cancelable = self.sendUnary { response in
-                        continuation.resume(returning: response)
+        await withTaskCancellationHandler {
+            await withCheckedContinuation { continuation in
+                guard !Task.isCancelled else {
+                    continuation.resume(returning: .init(code: .canceled, result: .failure(.canceled())))
+                    return
+                }
+
+                let hasResumed = Locked(false)
+                self.cancelable = self.sendUnary { response in
+                    guard !hasResumed.value else {
+                        return assertionFailure("Attempting to resume continuation twice")
                     }
+                    continuation.resume(returning: response)
+                    hasResumed.perform(action: { $0 = true })
                 }
             }
-        }, onCancel: {
-            Task { await self.cancelable?.cancel() }
-        })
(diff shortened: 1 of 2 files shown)


eseay:
@rebello95 I am going to keep this in Draft until I give the reporters from #333 the opportunity to integrate this branch and validate in some way. My means of reproducing the issue is so intermittent that it's honestly hard for me to tell whether my solution worked or if the problem is just not occurring because the reproduction steps are unsuccessful.

Libraries/Connect/Internal/Unary/UnaryAsyncWrapper.swift
```diff
@@ -21,7 +21,7 @@ import SwiftProtobuf
 /// https://forums.swift.org/t/how-to-use-withtaskcancellationhandler-properly/54341/37
 /// https://stackoverflow.com/q/71898080
 @available(iOS 13, *)
-actor UnaryAsyncWrapper<Output: ProtobufMessage>: Sendable {
+actor UnaryAsyncWrapper<Output: ProtobufMessage> {
```
rebello95, line 24: I'm curious as to the reason for this change? I believe it had to be `Sendable` before to prevent some warnings, but could be mistaken

Libraries/Connect/Internal/Unary/UnaryAsyncWrapper.swift
```diff
@@ -41,20 +41,33 @@ actor UnaryAsyncWrapper<Output: ProtobufMessage>: Sendable {
...
-                } else {
-                    self.cancelable = self.sendUnary { response in
-                        continuation.resume(returning: response)
+                    return
+                }
+
+                let hasResumed = Locked(false)
+                self.cancelable = self.sendUnary { response in
+                    guard !hasResumed.value else {
+                        return assertionFailure("Attempting to resume continuation twice")
```
rebello95, line 59: What's your thought process for keeping this assertion in place? I assume it will trigger in the cases described by the PR description, which could be disruptive for local development

rebello95 (commented):
Thank you for this fix! Looks good overall, just a couple questions

Libraries/Connect/Internal/Unary/UnaryAsyncWrapper.swift
eseay: This was actually something that SwiftLint raised up. Since it's an actor, it's already Sendable, so marking it as such was redundant.

rebello95: Oh nice 👍🏽 maybe we should update SwiftLint or opt-in to that rule on the repo then?

eseay: Good point. Adjusted accordingly.

Libraries/Connect/Internal/Unary/UnaryAsyncWrapper.swift
eseay: My reasoning was mostly that I felt like failing silently didn't really seem appropriate either.

In the interest of time, the change here specifically addresses the continuation resuming twice, which was resulting in the worst possible experience in both debug and release - a crash. It does not necessarily address the root cause for _why_ `sendUnary`'s closure is getting triggered a second time.

It felt like keeping an assertion there, such that the debugger pauses when it happens during development, would provide a means to re-raise the issue and prompt deeper analysis of the root cause at a later date if we see it happening regularly.

Up to you how you'd like to proceed though. I can just as easily `os_log` something, create an issue and put it in a comment, etc.

rebello95: > It felt like keeping an assertion there, such that the debugger pauses when it happens during development, would provide a means to re-raise the issue and prompt deeper analysis of the root cause at a later date if we see it happening regularly.
>
> Up to you how you'd like to proceed though. I can just as easily os_log something, create an issue and put it in a comment, etc.

I get where you're coming from. That said, if I'm understanding correctly this is an issue in the underlying library/dependency - not in how the caller is using the API. If a user hits an assertion in debug I think they'd think _they_ were doing something wrong, but in practice it doesn't seem like they need to do anything. Since they end up getting a non-successful response once anyway, I think adding an assertion will introduce confusion/distraction to the end user. IMO keeping an `os_log` here seems sufficient unless you feel strongly

eseay: Works for me! Would you also like us to make an Issue just to track the underlying problem for later investigation?

rebello95: I think an inline comment is probably sufficient :)

rebello95 (approved):
Thanks!

Issue #346, opened 2025-05-12
HTTP2StreamMultiplexer Sendable Conformance Issue in NIOHTTPClient


Leonif:
The error "Conformance of 'HTTP2StreamMultiplexer' to 'Sendable' is unavailable" is a Swift concurrency issue occurring with the NIO HTTP client. This error appears when the HTTP2StreamMultiplexer is used in a context where Swift's concurrency system expects a Sendable type.

```
self.state = .connecting
self.bootstrap
    .connect(host: self.host, port: self.port)
    .flatMap { channel -> EventLoopFuture<(Channel, HTTP2StreamMultiplexer)> in
        return .init(channel, HTTP2StreamMultiplexer) // The error occurs on this line
    }
    .whenComplete { ... }
```
The HTTP2StreamMultiplexer is not marked as Sendable, but it's being used in a context where Sendable conformance is required.

I use Xcode 16.2
Swift 6.0.3

rebello95:
Thanks for reporting this. Fixing in https://github.com/connectrpc/connect-swift/pull/347

PR #347, opened 2025-05-16
Fix compilation errors when using NIO

rebello95:
Resolves https://github.com/connectrpc/connect-swift/issues/346. Also updates the example app's `Package.resolved` (which updated itself when I opened the project).


rebello95:
Weird - conformance tests pass locally for me and the Eliza app works as well. @jhump or @eseay are either of you able to repro these errors locally?

eseay (approved):

PR #348, opened 2025-05-23
Adjust CI workflow definitions

eseay:
**Primary change:** Updates our CI jobs to use Xcode 16.3

**Secondary changes:**
- Adjusts the way that we specify our Xcode version to reduce duplication (and instances where changes need to be made in the future)
- Removes redundant installation of `xcbeautify` now that is [pre-installed on the runners](https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md#tools) (it was not always pre-installed)
- Renamed all workflow file extensions to `yml` for consistency


rebello95 (approved):
nice, thanks!

PR #350, opened 2025-06-12
Add SwiftFormat and SwiftLint disable commands to generated code

### LINKED ISSUE #349: Add swift format ignore & disable swift lint for generated file
I noticed our swift format really wants to reformat these files. It would be nice to make the codegen include

```
// swift-format-ignore-file
// swiftlint:disable all
```

at the beginning of each file, similar to what swift-protobuf codegen library does. This would prevent them from being auto-formatted.

eseay:
This mirrors the behavior that exists in swift-protobuf. One can see examples of this and the pattern it directly mirrors in any `.pb.swift` file.

The output from updated generator is demonstrated in the regenerated files within the repository itself.

Fixes #349


rebello95 (approved):
thanks!

rebello95:
Also I think you meant to comment/close https://github.com/connectrpc/connect-swift/issues/349 in the description

eseay:
> Also I think you meant to comment/close #349 in the description

🤦‍♂️ I couldn't figure out why it wasn't linking. It was because I fat-fingered the branch name to begin with. Good catch.

Issue #353, opened 2025-06-17
ConnectUnaryChannelHandler causes retain cycle with ChannelHandlerContext


Code-Hex:
### Summary  
`ConnectUnaryChannelHandler` keeps a strong reference to its `ChannelHandlerContext`, while the
context already holds a strong reference back to the handler.  
This two-way ownership forms a retain cycle, so neither object is deallocated after the request
finishes. Xcode’s Memory Graph shows a **2-node cycle** and `deinit` is never called.

### Environment
| Item | Value |
|------|-------|
| connect-swift | 1.0.4 |
| Swift | 6.1.2 (Xcode 16.4) |
| Platform | iOS 18.5 |

### Reproduction steps

1. Create a minimal client that makes repeated unary calls:

```swift
   import Connect
   import ConnectNIO
   import NIO

   let client = ProtocolClient(
            httpClient: NIOHTTPClient(host: build.proxyEndpointURL()),
            config: ProtocolClientConfig(
                host: build.proxyEndpointURL(),
                networkProtocol: .connect,
                codec: ProtoCodec()
            )
        )

// Use client
```

2. Run in Xcode with **Debug Memory Graph**.
3. After the loop completes, open the memory graph:

   * `ConnectUnaryChannelHandler` instances remain alive
   * Each retains its `ChannelHandlerContext`
   * The context retains the handler ⇒ 2-node cycle.

![Image](https://github.com/user-attachments/assets/2779c818-da94-49f4-a3ee-14bca45d7811) 

![Image](https://github.com/user-attachments/assets/00ecc055-9ad9-4fef-8d8e-7a2a51cd0830)

### Root cause (probably)

`ConnectUnaryChannelHandler` declares

```swift
private var context: ChannelHandlerContext? // strong reference
```

`ChannelHandlerContext` already stores a strong reference to its handler, so the two objects
retain each other indefinitely once the request completes.

### Proposed fix (Option A – simplest)

```diff
- private var context: ChannelHandlerContext?
+ private weak var context: ChannelHandlerContext?
```

### Proposed fix (Option B – if `weak` is undesired)

Explicitly break the cycle when the channel closes:

```swift
func channelInactive(context: ChannelHandlerContext) {
    self.context = nil
    context.fireChannelInactive()
}

func handlerRemoved(context: ChannelHandlerContext) {
    self.context = nil
}
```

Either change removes the retain cycle; `deinit` is invoked and the memory graph is clean.

### Additional notes

[...]

PR #354, opened 2025-06-18
fixed memory leak caused by retain cycle in channel handlers

Code-Hex:
fixes: https://github.com/connectrpc/connect-swift/issues/353


rebello95:
Hmm trying re-running just in case it's a flake

rebello95 (approved):
thank you for the fix!

PR #355, opened 2025-06-28
Add package visibility option to code generators

eseay:
The `package` access modifier was introduced in Swift 5.9 ([proposal 0386](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0386-package-access-modifier.md)) to bridge the gap between `internal` and `public` in Swift packages.

[Support for this modifier in generated code](https://github.com/apple/swift-protobuf/pull/1472) was added to SwiftProtobuf in September 2023.

This change adds the `Package` visibility option to the Connect generators. This is particularly useful when developing a Swift package, and you want to vend your generated models and mocks from a specific target to other targets of your package but not to dependent applications.


eseay:
@rebello95 requesting second review since I made a change since your approval. I recalled that conformance was for some reason failing on a previous PR due to something related to an outdated Xcode version, so I updated Xcode here, and it was happy again.

Issue #358, opened 2025-07-04
♻️ Memory leak when ending/restarting streaming calls (Connect Swift)


patryk-sredzinski:
We've observed a memory leak in `ConnectRPC` when streaming calls are repeatedly started and stopped. The issue reproduces with the official example app.

---

### ✅ Steps to Reproduce:

1. Clone the Connect Swift repo  
2. Open:  
   `Examples/ElizaSwiftPackageApp/ElizaSwiftPackageApp.xcodeproj`  
3. Run the app on the iOS Simulator  
4. Tap **Connect (Streaming)**  
5. Tap **End Chat**  
6. Repeat steps 4–5 multiple times  
7. Open **Memory Graph Debugger** in Xcode  

---

### 🐛 Observed:

- Each new stream adds retained objects that are never released  
- Over time, memory usage grows  
- Objects are retained even after ending the stream  

---

### 📹 Reproduction Video:

https://youtu.be/WmKnDdT4jzs

---

### 💡 Expected:

- Stream resources should be properly released
- No increasing memory footprint with repeated sessions  

---

### 🔧 Environment:

- Connect Swift version: 1.1.0
- Xcode: 16.4
- iOS Simulator: iOS 18.5
- Reproduced on a clean build of the example app

PR #359, opened 2025-07-04
Weakify urlSessionStream in request callbacks to avoid retain cycles

patryk-sredzinski:
This is a proposal that solves the issue: #358


rebello95 (approved):
Thank you for the fix! Please address the DCO CI failure, and we can merge after 😄

rebello95:
Yes, thank you!

PR #368, opened 2025-09-20
Compression Query Paramater Fix

cnoon:
This PR resolves an issue we found with `unaryGET` requests between iOS and Android libraries when looking at our CDN logs. 

## Background
We noticed that we were appending `compression=identity` to all our iOS requests and not our equivalent Android requests. That led me on quite a journey to understand how this all works. The [connect-kotlin](https://github.com/connectrpc/connect-kotlin/blob/01db4796e47bf5f2a4b677ec1ea6f483a2344bd7/library/src/main/kotlin/com/connectrpc/protocols/ConnectInterceptor.kt#L362-L364) and [connect-swift](https://github.com/connectrpc/connect-swift/blob/main/Libraries/Connect/Internal/Interceptors/ConnectInterceptor.swift#L250-L260) implementations are not currently the same, and the changes proposed in this PR bring them as close together as possible without a much larger refactor to the iOS approach.

Screenshot 2025-09-19 at 5 46 40 PM

## Proposal
Prior to this change, the `unaryGET` logic was incorrectly always appending the `compression=identity` query parameter even when compression wasn't being applied. This logic has now been modified ( with ✅ tests ✅ added 😎 ) to only apply the query parameter when the `Content-Type` header has been set. You cannot use the `requestCompression` object attached to the `ProtocolClientConfig` since the data has already been compressed when it reaches this part of the code. Because of this, I opted to use the `Content-Type` header as the signal to apply the query parameter.

Please have a deeper look at this to see if you'd rather use the same approach as the Kotlin library does. I think it would be better to use the `requestCompression` object to determine whether it should be set or not, but that will require an order of operations refactor to this system.

Cheers. 🍻

### DIFF
--- a/Tests/UnitTests/ConnectLibraryTests/ConnectTests/ProtocolClientConfigTests.swift
+++ b/Tests/UnitTests/ConnectLibraryTests/ConnectTests/ProtocolClientConfigTests.swift
@@ -214,7 +214,7 @@ final class ProtocolClientConfigTests: XCTestCase {
         ).shouldUseUnaryGET(for: request))
     }
 
-    func testTransformToGETIfNeededWithRequestCompression() {
+    func testTransformToGETWithoutRequestCompression() {
         let request = HTTPRequest<Data?>(
             url: URL(string: "https://connectrpc.com")!,
             headers: Headers(),
@@ -237,7 +237,9 @@ final class ProtocolClientConfigTests: XCTestCase {
             requestWithoutCompression.url.absoluteString,
             "https://connectrpc.com?base64=1&connect=v1&encoding=json&message=AAEC"
         )
+    }
 
+    func testTransformToGETWithRequestCompression() {
         let compressedRequest = HTTPRequest<Data?>(
             url: URL(string: "https://connectrpc.com")!,
             headers: [HeaderConstants.contentEncoding: ["gzip"]], // mimics inceptor behavior
@@ -245,6 +247,12 @@ final class ProtocolClientConfigTests: XCTestCase {
             method: .post,
             trailers: nil,
             idempotencyLevel: .noSideEffects
+        )
+
+        let config = ProtocolClientConfig(
+            host: "https://connectrpc.com",
+            networkProtocol: .connect,
+            unaryGET: .alwaysEnabled
         )
 
         let requestWithCompression = config.transformToGETIfNeeded(compressedRequest)


Libraries/Connect/Internal/Interceptors/ConnectInterceptor.swift
```diff
@@ -241,23 +241,28 @@ extension ConnectInterceptor: StreamInterceptor {
     }
 }
 
-private extension ProtocolClientConfig {
+extension ProtocolClientConfig {
```
cnoon, line 244: Likely necessary to test without a large refactor to how the public APIs leverage this.

Tests/UnitTests/ConnectLibraryTests/ConnectTests/ProtocolClientConfigTests.swift
```diff
@@ -213,4 +213,46 @@ final class ProtocolClientConfigTests: XCTestCase {
             unaryGET: .disabled
         ).shouldUseUnaryGET(for: request))
     }
+
+    func testTransformToGETIfNeededWithRequestCompression() {
```
rebello95, line 217: I think it'd make sense for this to be split into 2 tests (one with compression and one without)

Libraries/Connect/Internal/Interceptors/ConnectInterceptor.swift
```diff
@@ -241,23 +241,28 @@ extension ConnectInterceptor: StreamInterceptor {
...
-            URLQueryItem(
-                name: "compression",
-                value: request.headers[HeaderConstants.contentEncoding]?.first ?? "identity"
-            ),
             URLQueryItem(name: "connect", value: "v\(ConnectInterceptor.protocolVersion)"),
             URLQueryItem(name: "encoding", value: self.codec.name()),
             URLQueryItem(name: "message", value: request.message?.base64EncodedString()),
         ]
+
+        if let contentEncoding = request.headers[HeaderConstants.contentEncoding]?.first {
```
rebello95, line 259: @jhump I wonder if there's an opportunity to add a conformance test for this?

rebello95 (approved):
Overall this seems reasonable to me, thank you for the contribution! I have one comment and would like @jhump to take a look at this as well since he's a maintainer of connect-kotlin

Libraries/Connect/Internal/Interceptors/ConnectInterceptor.swift
jhump: I'm trying to parse the PR description to understand the issue.

> incorrectly always appending the compression=identity query parameter even when compression wasn't being applied.

Did this intend to say "even when compression WAS being applied"? Is the issue that it reports "compression=identity" in a query param but was actually compressing the payload?

If so, yeah. Looks like the conformance suite has a TODO about this actually in the definition of the "Connect with GET" test suite:
```
  # TODO: The reference client only uses compression for GET requests if the
  # request is larger than a particular size. So since we aren't sending a
  # request here that exceeds that, our requests won't use compression even
  # if we specify it. So, we're limiting this test to identity.
```
I am pretty certain we can control that size threshold via a client option and can likely set it to zero (or one; something arbitrarily small), so it always uses compression when we want/expect it to do so.

cnoon: No @jhump. That phrase is actually written correctly. Right now, no matter what, this library is always appending `compression=identity` query parameters when there is no compression being applied. We have number of GET RPCs that we're using that have very small request payloads which were never coming close to the request compression limits we had put in place, but we were still seeing that query parameter on iOS and not Android builds. After digging much deeper, we realized that iOS was just always appending `compression=identity` until the actual compression started to kick in.

rebello95:
Going to merge this despite the watchOS failures - those will be fixed by https://github.com/connectrpc/connect-swift/pull/370

Libraries/Connect/Internal/Interceptors/ConnectInterceptor.swift
jhump: > we realized that iOS was just always appending compression=identity until the actual compression started to kick in

Okay. Is that a bug? A `compression=identity` parameter basically means "no compression". So this client is definitely conforming to the spec as is. If you want to elide the parameter, just to make the entire request URI shorter, that is an implementation decision. I suppose an argument could be made that we should have a conformance test for this, in order to enforce "best practices". But we don't really have a good way (at least not yet) in the conformance suite of marking some test cases as "required" and others as "recommended". This particular assertion would definitely be recommended, not required.

PR #370, opened 2025-10-01
Update CI runners

rebello95:
Updating to the latest since it looks like some jobs are failing on old simulators.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,7 @@ permissions:
   contents: write
 jobs:
   release:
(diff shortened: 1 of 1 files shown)


eseay:
@rebello95 The OS and Xcode bumps look good!

Two other things that are directly related to those bumps that may be worthwhile to just knock out at the same time:
- We can bump the runners to `macos-26`
- We could also up the simulator names from "iPhone 16" to "iPhone 17" for long term compatibility with the corresponding Xcode and macOS version increments.

eseay (approved):

PR #374, opened 2025-11-25
Fix CI failures resulting from missing simulators

rebello95:
GitHub Actions keeps changing supported simulators on our CI machines from underneath us, breaking CI checks (example here: https://github.com/connectrpc/connect-swift/actions/runs/19645507500/job/56271087929?pr=373).

This PR removes the simulator/OS pinning - I think this will result in fewer net CI failures since we can generally continue using the defaults on the machines (note that we are still pinning macOS versions on the machines, as well as Xcode versions).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - name: Build Connect iOS library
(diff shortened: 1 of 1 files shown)


eseay:
It looks like the issue is that we still need to pin the OS version, but the device pinning should be fine to remove. The OS version is somewhat tied to the Xcode version we're targeting too, so keeping those updated and in sync may just have to be something we deal with.

rebello95:
Going with https://github.com/connectrpc/connect-swift/pull/375 instead

PR #376, opened 2025-11-27
Fix unreliable timeout

### LINKED ISSUE #361: Unreliable timeout in 1.1.0
I recently updated the Connect lib to `1.1.0` in an app, and was able to notice that the timeout specified in `ProtocolClient` and `ProtocolClientConfig` seems to not be working reliably. Are you aware of this?

I can't reproduce every time, but was still able to do it multiple times + had user feedback about it.

### Step to reproduce:

- Set a timeout
- Set Network Link Conditioner on, with `100% Loss` config
- Trigger the request

Image

### Outcome:
- The request never gets an answer, and Connect never didn't time out either.
- If you are awaiting for the result in a specific loading state, the app get stuck in this state

### Expected behavior:
- The timeout fires consistently if no response is received after the specified time.

eLud:
### Problem

  When network connectivity is lost (e.g., 100% packet loss), requests hang indefinitely instead of timing out or returning an error. This is a regression
  introduced in v1.1.0.

  Fixes #361

  ### Root Cause

  When network drops, NIO's HTTP/2 layer detects the connection loss and triggers `channelInactive` before the `IdleStateHandler` can fire its timeout event.
   However, `channelInactive` only closed the connection without notifying the caller:

  ```swift
  func channelInactive(context: ChannelHandlerContext) {
      self.closeConnection()
      context.fireChannelInactive()
      // ❌ onResponse was never called - request hangs forever
  }
```

  This worked in v1.0.4 because the retain cycle (fixed in #354) kept handlers alive longer, giving IdleStateHandler time to fire. With the fix, cleanup happens faster and `channelInactive` wins the race—but doesn't notify the caller.

 ### Solution

  1. Added a `hasResponded` flag to track whether the response callback has been invoked
  2. Modified `channelInactive` to call the response callback with `.unavailable` error if no response was received yet

```swift
  func channelInactive(context: ChannelHandlerContext) {
      let shouldNotify = !self.hasResponded
      self.closeConnection()
      if shouldNotify {
          self.hasResponded = true
          self.onResponse(.init(
              code: .unavailable,
              error: ConnectError(code: .unavailable, message: "Channel became inactive", ...)
          ))
      }
      context.fireChannelInactive()
  }
```

 ### Changes

  - `ConnectUnaryChannelHandler.swift`: Added `hasResponded` flag and updated `channelInactive`
  - `ConnectStreamChannelHandler.swift`: Added `hasResponded` flag and updated `channelInactive`

 ### Testing

  Tested with Network Link Conditioner at 100% packet loss:
  - **Before:** Request hangs indefinitely
  - **After:** Request returns .unavailable error promptly


rebello95:
Thanks for putting up this potential fix! Once https://github.com/connectrpc/connect-swift/pull/378 lands (fixes some CI issues), would you mind updating this branch so that we can run the conformance test suite against it on CI? Looks like you'll also have to update your commit to pass DCO

Libraries/ConnectNIO/Internal/ConnectUnaryChannelHandler.swift
```diff
@@ -51,6 +52,7 @@ final class ConnectUnaryChannelHandler: NIOCore.ChannelInboundHandler, @unchecke
             }
 
             self.closeConnection()
+            self.hasResponded = true
```
rebello95: Alternatively to my previous comment, what do you think about moving `self.hasResponded = true` to `closeConnection`? Seems like that would simplify this a bit

rebello95:
Sounds good, thanks @eLud!

rebello95:
> Hum, tested the conformance tests, and they do pass… I updated my branch to have latest changes from main too. Could you re-run the CI to check @rebello95?

Thanks for trying. I ran CI a few more times but they're still failing 🤔 @eseay or @jhump do either of you have ideas on this one?

PR #379, opened 2025-12-05
Upgrade conformance suite to v1.0.5

rebello95:
Upgrading to https://github.com/connectrpc/conformance/releases/tag/v1.0.5, and bumping CI to use Go 1.24.x which it requires.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
-          go-version: 1.23.x
+          go-version: 1.24.x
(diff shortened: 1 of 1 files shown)


rebello95:
Seeing some flakes on CI runs on this PR which I _think_ may be fixed by https://github.com/connectrpc/connect-swift/pull/376

jhump:
> bumping CI to use Go 1.24.x which it requires

Ah, interesting that this repo is building it from source. I guess I could put a note in future release notes about the Go requirements changes.

FWIW, the connect-kotlin repo downloads prebuilt artifacts for the GitHub release instead of building from source:
https://github.com/connectrpc/connect-kotlin/blob/main/Makefile#L107-L120

PR #397, opened 2026-06-01
Use dedicated serial OperationQueue instead of main queue for URLSession delegate callbacks

chengmin-z:
`URLSessionHTTPClient` was previously initialized with delegateQueue: `.main`, causing all URLSession delegate callbacks (response handling, data delivery, task completion, metrics collection) to run on the main thread. 
This could block the UI when performing high-throughput or long-running network requests.

This change replaces the main queue with a dedicated OperationQueue that:

- Runs on a background thread with .userInitiated QoS

- Limits concurrency to 1 so that delegate callbacks remain ordered and serialized

- Uses a reverse-DNS name for easier debugging in Instruments and crash reports

⚠️ Breaking Change

1. ProtocolClient defaults to URLSessionHTTPClient — callers using ProtocolClient(config:) without a custom httpClient will see delegate callbacks on a background queue.
2. Subclasses of URLSessionHTTPClient that assumed delegate methods run on the main thread may break.

The Connect library does not guarantee main-thread delegate callbacks. If you need main thread execution, dispatch to DispatchQueue.main in your own code.

### DIFF
--- a/Libraries/Connect/Public/Implementation/Clients/URLSessionHTTPClient.swift
+++ b/Libraries/Connect/Public/Implementation/Clients/URLSessionHTTPClient.swift
@@ -34,11 +34,15 @@ open class URLSessionHTTPClient: NSObject, HTTPClientInterface, @unchecked Senda
     private var streams = [Int: URLSessionStream]()
 
     public init(configuration: URLSessionConfiguration = .default) {
+        let delegateQueue = OperationQueue()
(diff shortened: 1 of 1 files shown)


rebello95 (approved):
seems reasonable to me. thanks!

chengmin-z:
> seems reasonable to me. thanks!

Awesome, glad we’re on the same page. We just bumped into this issue using Protobuf in our app.

PR #399, opened 2026-06-01
Fix crash when URLSession resends a streaming request body

### LINKED ISSUE #398: SIGABRT in HTTPRequestBody.cpp when URLSession resends a streaming request body (needNewBodyStream called twice)
## Summary

Long-lived streaming RPCs over `URLSessionHTTPClient` can crash the process with:

```
SIGABRT: _onqueue_setupStream_block_invoke > Assertion failed:
(CFReadStreamGetStatus(_stream.get()) == kCFStreamStatusNotOpen),
function _onqueue_setupStream_block_invoke, file HTTPRequestBody.cpp
```

We see this in production on iOS 15–17 for `application/connect` server-streaming calls. The crashes correlate strongly with **background → foreground transitions** (i.e. the network path changes and `URLSession` transparently retries the in-flight streaming request).

## Root cause

`URLSessionHTTPClient` implements `urlSession(_:task:needNewBodyStream:)` by returning the same bound `InputStream` every time:

```swift
open func urlSession(
    _ session: URLSession, task: URLSessionTask,
    needNewBodyStream completionHandler: @escaping (InputStream?) -> Void
) {
    completionHandler(
        self.lock.perform { self.streams[task.taskIdentifier]?.requestBodyStream }
    )
}
```

…where `URLSessionStream.requestBodyStream` always returns the single bound `readStream`.

Per [Apple's documentation](https://developer.apple.com/documentation/foundation/urlsessiontaskdelegate/urlsession(_:task:neednewbodystream:)), this delegate method is called under **two** circumstances:

> - To provide the initial request body stream if the task was created with `uploadTask(withStreamedRequest:)`
> - To provide a **replacement** request body stream if the task needs to **resend** a request that has a body stream because of an authentication challenge or other recoverable server error.

On the second (resend) call, `URLSession` is handed an `InputStream` that has **already been opened** by the first attempt. CFNetwork asserts the stream must be `kCFStreamStatusNotOpen` and aborts the process. Apple's own streaming-upload sample notes that when a body stream fails you must "close the streams and abandon the upload" — i.e. a streamed body cannot be replayed.

## Expected behavior

Because a streamed request body cannot be replayed, the client should not vend the already-consumed stream a second time. Returning `nil` on the resend fails the task cleanly (surfaced through the normal stream-close path), allowing callers to reconnect instead of crashing.

## Environment

- connect-swift `1.2.2` (also present on `main`)
- `networkProtocol: .connect`, `URLSessionHTTPClient`
- Server streaming, iOS 15–17

## Proposed fix

Vend the bound input stream only once in `URLSessionStream`, returning `nil` on subsequent requests. PR incoming.

patryk-sredzinski:
## Summary

`urlSession(_:task:needNewBodyStream:)` is called by `URLSession` both to obtain the initial body stream for a streamed upload task **and** to resend a request that has a body stream after a recoverable error (per [Apple's docs](https://developer.apple.com/documentation/foundation/urlsessiontaskdelegate/urlsession(_:task:neednewbodystream:))). In practice the resend path is hit when an in-flight streaming connection drops and is retried — commonly on a background → foreground transition when the network path changes.

`URLSessionStream` always returned the same bound `InputStream`. On the resend, `URLSession` receives an input stream that has already been opened, and CFNetwork aborts the process:

```
Assertion failed: (CFReadStreamGetStatus(_stream.get()) == kCFStreamStatusNotOpen),
function _onqueue_setupStream_block_invoke, file HTTPRequestBody.cpp
```

A streamed request body cannot be replayed, so this vends the bound input stream **only once**. Subsequent requests return `nil`, which fails the task and is surfaced through the normal stream-close path instead of crashing. (`URLSessionHTTPClient.needNewBodyStream` already forwards the value via optional chaining, so no change is needed there.)

Fixes #398.

## Changes

- `URLSessionStream.requestBodyStream` is now `InputStream?` and vends the bound stream only once (guarded by a `Locked` flag), returning `nil` thereafter.
- Adds `URLSessionStreamTests.testVendsRequestBodyStreamOnlyOnce`.

## Testing

- `swift build --target Connect`
- `swift test --filter URLSessionStreamTests` (passes)
- `swiftlint lint` clean on changed files

## Notes

- Backward compatible: `requestBodyStream` is `internal`, so the visible behavior change is only that a streamed request body is no longer replayed (which previously crashed).

### DIFF
--- a/Libraries/Connect/Internal/Streaming/URLSessionStream.swift
+++ b/Libraries/Connect/Internal/Streaming/URLSessionStream.swift
@@ -21,6 +21,7 @@ import Foundation
 final class URLSessionStream: NSObject, @unchecked Sendable {
     private let closedByServer = Locked(false)
     private let readStream: Foundation.InputStream
+    private let requestBodyStreamVended = Locked(false)
     private let responseCallbacks: ResponseCallbacks
     private let task: URLSessionUploadTask
     private let writeStream: Foundation.OutputStream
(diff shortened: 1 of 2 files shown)


rebello95 (approved):
thanks for the fix!

PR #404, opened 2026-07-19
Migrate unit tests from XCTest to Swift Testing

eseay:
**Contribution guide:** read and followed per https://github.com/connectrpc/connect-swift/blob/main/.github/CONTRIBUTING.md

## Summary

Migrates all 15 unit test suites under `Tests/UnitTests` (`ConnectLibraryTests` and `ConnectPluginUtilitiesTests`) from XCTest to [Swift Testing](https://developer.apple.com/documentation/testing), one suite per commit for reviewability. No `Package.swift`, CI, or Makefile changes were needed — `swift test` discovers Swift Testing tests as-is.

**Mechanical conversions:** `XCTestCase` classes → struct suites; `XCTAssert*` → `#expect`; `XCTUnwrap` → `#require`; `XCTFail` → `Issue.record`; `XCTAssertThrowsError` → typed `#expect(throws:)`.

**Notable decisions:**
- The four near-identical `FilePathComponentsTests` methods are collapsed into one parameterized `@Test(arguments:)` (72 original methods → 69 discovered tests; no coverage lost).
- Each `@Test` function carries `@available(iOS 13, *)` because the package's iOS floor is 12, below Swift Testing's iOS 13 minimum. The annotation must be per-function: the `@Test` macro rejects test functions inside `@available` types.
- `InterceptorIntegrationTests` is `@Suite(.serialized)` (documented in-source): Swift Testing parallelizes by default, and these tests share the live conformance reference server on port 52107 and assert timing-sensitive URLSession metrics. Its blocking `sleep(1)` is now a cooperative `Task.sleep`. The iOS 16+ metrics tests also carry a one-minute `.timeLimit` so a hung reference server fails CI instead of hanging it.

**Lint changes (`.swiftlint.yml`):** removes the now-dead XCTest rules (`empty_xctest_method`, `xct_specific_matcher`, custom `xctestcase`). No Swift Testing–specific lint rules are added; `@Test` attributes follow the existing `attributes` opt-in rule (own line above the function).

## Test plan

- [x] `swift test` via `make testunit` (against the conformance reference server): 69 tests in 15 suites pass.
- [x] `xcodebuild test` on iOS Simulator (iPhone 17 Pro) with the reference server running: all tests pass, confirming the iOS 12 deployment target compiles with the availability annotations.
- [x] `make testconformance`: 966/966 (URLSession) and 1681/1681 (NIO) conformance cases pass.
- [x] `swiftlint lint --strict`: 0 violations.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -117,3 +117,11 @@ custom_rules:
   spaces_over_tabs:
     name: "Use (4) spaces instead of tabs"
     regex: '\t'
+  swift_testing_over_xctest:
+    name: "Use Swift Testing rather than XCTest for writing tests"
+    message: "Import Testing and write tests with @Test/#expect instead of XCTest."
+    regex: '^\s*(@testable\s+)?import\s+XCTest\b'
+    match_kinds:
+      - attribute.builtin
+      - keyword
+      - identifier
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -120,8 +120,4 @@ custom_rules:
   swift_testing_over_xctest:
     name: "Use Swift Testing rather than XCTest for writing tests"
     message: "Import Testing and write tests with @Test/#expect instead of XCTest."
-    regex: '^\s*(@testable\s+)?import\s+XCTest\b'
-    match_kinds:
-      - attribute.builtin
-      - keyword
-      - identifier
+    regex: 'XCTest'
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -35,7 +35,6 @@ opt_in_rules:
   - empty_collection_literal
   - empty_count
   - empty_string
-  - empty_xctest_method
   - enum_case_associated_values_count
   - explicit_init
   - fallthrough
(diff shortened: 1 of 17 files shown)


.swiftlint.yml
```diff
@@ -119,6 +117,3 @@ custom_rules:
   spaces_over_tabs:
     name: "Use (4) spaces instead of tabs"
     regex: '\t'
```
rebello95, line 119: Should we lint against XCTest in general in favor of Swift testing?

rebello95 (approved):
thanks!

.swiftlint.yml
eseay: I think that's a good idea. Let me add that.

.swiftlint.yml
```diff
@@ -119,6 +117,11 @@ custom_rules:
...
-  xctestcase:
-    name: "Use XCTestCase over XCTest to ensure tests run properly"
-    regex: ': XCTest[\s,]+'
+  swift_testing_over_xctest:
```
rebello95, line 120: Would imply linting against XCTest work?

rebello95 (approved):

.swiftlint.yml
eseay: It technically would. This is a bit "smarter," but it's overkill.

rebello95 (approved):

PR #406, opened 2026-07-20
Swift 6 strict-concurrency migration

eseay:
## Summary

Completes the Swift 6 / strict-concurrency migration. The package already compiled in Swift 6
language mode (via `swiftLanguageVersions`), so this work makes that contract explicit, drives the
remaining data-race-safety warnings to zero, and triages every concurrency escape hatch.

**Result:** the package builds with **zero warnings** under
`swift build --build-tests -Xswiftc -warnings-as-errors` (libraries and tests), now enforced in CI.

## What changed

- **Manifest:** `swift-tools-version` 5.6 → 6.0; `swiftLanguageVersions` → `swiftLanguageModes: [.v6, .v5]`. `.v5` is retained to document the dual-mode contract for CocoaPods consumers.
- **`@unchecked Sendable` triage:** 3 sites removed by restructuring to checked `Sendable` (`TimeoutTimer`, `ClientOnlyStream`, `PendingRequestCallbacks`); every remaining site now carries an inline safety-invariant + removal-plan comment.
  - Moving those types' previously-unsynchronized `var`s behind `Locked` also closes real races — `BidirectionalAsyncStream.requestCallbacks` was written by `configureForSending` on the caller's thread and read by `AsyncStream.onTermination` on arbitrary threads.
- **`ConnectNIO` channel handlers:** `ConnectUnaryChannelHandler` and `ConnectStreamChannelHandler` become checked `Sendable`. Their loop-confined mutable state collapses into a single `State` struct held in a `NIOLoopBoundBox`, which runtime-asserts that every read/write happens on the handler's event loop. `startMultiplexChannel` now constructs the non-Sendable codec/idle-state handlers inside the stream-channel initializer — capturing only `useSSL`/`timeout` — instead of reaching back through `self`.
- **Podspecs:** `swift_versions = ['5.0', '6.0']` so CocoaPods consumers build in Swift 6 mode on modern Xcode, with `'5.0'` retained as an escape hatch.
- **CI:** new `build-strict-concurrency` job fails on any reintroduced warning.
- **README:** documents the Xcode 16 / Swift 6 floor for SwiftPM consumption.

### Fixes from adversarial review of this branch

- **`TimeoutTimer` cancelation is sticky again.** An earlier commit on this branch moved `DispatchWorkItem` creation out of `init` and into `start()`, which left `cancel()` a no-op whenever it ran before the item existed — `start()` would then arm a fresh, uncancelled timer. The original design relied on the item living for the object's whole lifetime, so a cancel-before-start permanently disarmed the timer (a cancelled `DispatchWorkItem` submitted via `asyncAfter` is never invoked). That window is reached in practice: `ProtocolClient` hands the transport a terminal callback that calls `timeoutTimer?.cancel()` *before* it reaches `timeoutTimer?.start(...)`, so a fast failure — disconnected NIO client, immediate URLSession error — lost the cancelation and a stale timer later fired against a finished request. `isCanceled` and `workItem` now live under one lock, so cancelation is sticky and atomic with arming. A 100-timer concurrent cancel/start harness lost **58 cancelations before the fix and 0 after**; both that case and cancel-before-start are now regression tests in `TimeoutTests.swift`.
- **`NIOHTTPClient`'s event loop group constraint is documented.** The handlers bind their `NIOLoopBoundBox` to `loopGroup.next()`, and `loopGroup` is `private`, so a subclass cannot build a bootstrap on it. A from-scratch `createBootstrap()` override therefore puts channel callbacks on a foreign loop and trips the box's precondition, which is live in release builds — a silent data race before this branch, a hard trap after it. `createBootstrap()` now documents the constraint and points overrides at `super.createBootstrap()` (which preserves the group and is unaffected), and a debug `assert` on connect names the cause instead of trapping deep inside NIO.
- **`ClientOnlyAsyncStream`'s safety comment is corrected.** It described the restated `@unchecked Sendable` as "redundant but harmless"; it is in fact required — removing it fails to compile with *"must restate inherited '@unchecked Sendable' conformance"*, because Swift makes each subclass re-assert the unchecked invariant for the state it adds.

## Context — already on `main`, *not* in this diff

- The `NIOHTTPClient.deinit` crash (`syncShutdownGracefully()` traps when `deinit` runs on the group's own event loop) was the root cause of the flaky conformance "timed out waiting for result" bursts. It landed separately in #407. Worth restating here because Swift 6 **cannot** catch that class of bug — `deinit` isolation is unchecked — so the migration would never have surfaced it.
- The `syncOperations`-based bootstrap pipeline setup, and the removal of the `@preconcurrency` NIO imports, landed in #411.

## Compatibility notes

- **No public API changes.** `URLSessionHTTPClient` and `NIOHTTPClient` stay `open` and `@unchecked Sendable`; `NIOHTTPClient.createBootstrap()` keeps its signature. Generated-code templates are unchanged (regeneration produces no diff).
- **The `ConnectNIO` handler rewrites are storage-only:** same control flow, same callbacks, same ordering.
- **No compiler-floor change beyond the manifest.** The library sources use no `nonisolated(unsafe)`, so the Swift 5 CocoaPods path builds on the same compilers as before. SwiftPM consumption requires Xcode 16+ solely because of `swift-tools-version:6.0`.
- Deployment targets unchanged (iOS 12 / macOS 10.15 / tvOS 13 / watchOS 6).

## Verification

Local, on Xcode 26.6 / Swift 6.3.3:

- `swift build --build-tests -Xswiftc -warnings-as-errors` — clean
- Swift 5 language mode (temporarily pinning `swiftLanguageModes: [.v5]`) — clean, validating the CocoaPods path
- `swift test` — 72/72 (requires the reference server on port 52107; `make testunit` starts it)
- `make testconformance` — 966 + 1681 cases, 0 failed, both `urlsession` and `nio`
- `swiftlint lint --strict` — clean
- Codegen no-diff, iOS/macOS builds, SwiftPM example

**Relying on CI to confirm** (not runnable locally): tvOS/watchOS builds, the CocoaPods example, and the warning inventory under CI's pinned Xcode.

## Known gaps

1. **The zero-warning gate is macOS-only.** `build-strict-concurrency` runs `swift build`, so `build-library-{ios,tvos,watchos}` still build without `-warnings-as-errors`. Platform-conditional code could reintroduce a strict-concurrency warning without failing CI. Deliberately left out of this PR.
2. **`NIOHTTPClient.loopGroup` is still `private`.** The constraint above is documented and asserted, but the complete fix is exposing the group so an override can actually use it — that is additive public API and out of scope here.
3. Conformance runs emit ~25 `Cannot schedule tasks on an EventLoop that has already shut down` errors. **Pre-existing** — `main` emits a comparable count — but NIO states this becomes a forced crash in a future release, so it deserves its own issue.


eseay:
Attaching migration plan so that it can be removed from the actual diff.

[SWIFT6_MIGRATION_PLAN.md](https://github.com/user-attachments/files/30281297/SWIFT6_MIGRATION_PLAN.md)

PR #407, opened 2026-07-20
Fix conformance test CI flakiness (NIO deinit crash + URLSession body-resend hang)

eseay:
## Summary

Fixes the long-standing flakiness in the `run-conformance-tests` CI job, where a run would emit a nameless `timed out waiting for result from client` line and then, ~5s later, a burst of 80–100 `FAILED: … failed to get result from client` lines spanning unrelated test categories.

**Root cause: the conformance client was crashing/hanging *between* test cases, not failing individual tests.** `Tests/ConformanceClient/Sources/main.swift` reads length-prefixed requests from stdin and runs them **serially** — one process, one test at a time. When the process dies or wedges between tests, the runner times out on the in-flight test and then fails every test still queued behind it at once. That is why the "failing" tests looked random and unrelated each run — they were never actually run.

Two distinct underlying bugs produced this same signature; both are fixed here.

### 1. `NIOHTTPClient` crash on deinit (SIGTRAP)

Reproduced under load with matching `.ips` crash reports. If the last strong reference to a `NIOHTTPClient` is dropped while an `EventLoopFuture` callback (e.g. the connect callback in `connectChannelAndMultiplexerIfNeeded()`) holds a temporary strong reference upgraded from `[weak self]`, the client is deallocated **on its own event loop thread**. `deinit` then called `syncShutdownGracefully()`, which traps with a precondition failure when invoked from an event loop — killing the process with SIGTRAP, silently, between test cases.

**Fix:** use the asynchronous `shutdownGracefully(_:)`, which is safe to call from any thread including the group's own event loops.

### 2. `URLSessionStream` hang on request-body resend

Reproduced locally and in CI (runs `29702874348`, `29702882058`). Every hung test was a streaming RPC over `URLSessionHTTPClient`; in one case even a 2-second `TimeoutTimer` calling `task.cancel()` could not produce a completion callback. When a connection is dropped and retried, URLSession asks for the request body again via `urlSession(_:task:needNewBodyStream:)`. A bound stream pair can't be replayed, so `requestBodyStream` vends `nil` (#399). Contrary to the assumption there, CFNetwork does **not** fail the task when given a `nil` body stream — it can leave the task parked indefinitely with no delegate callbacks, wedging the serial client.

**Fix:** cancel the task when a resend is requested, guaranteeing `didCompleteWithError` is delivered, and surface the failure as `.unavailable` with a descriptive message (rather than a misleading `canceled`) since it is a retriable connection-level failure.

### 3. Diagnosability

The conformance client now logs `starting <test> …` / `finished <test>` to stderr before/after each case. The runner surfaces client stderr, so a future wedge shows exactly which test/protocol/codec/streamType was in flight — the piece of information missing from every prior CI failure log, and what pinpointed both bugs here.

## Validation

- **NIO crash:** 1,071 suite runs with the fix while the machine was verifiably awake (534 nio / 537 urlsession) produced **zero NIO crashes and zero NIO bursts**; ~17 crashes were expected at the pre-fix rate. Unit tests: 69/69 pass.
- **URLSession hang:** validated on the real macos-26 CI runners (in progress on this branch).
- Note for reproduction: laptop sleep fires the runner's wall-clock timeouts and produces fake "burst" failures on wake — all local runs were cross-checked against `pmset -g log` and sleep-window runs discarded.

## Notes

This branch was previously stacked on #404 (Swift Testing migration); it has been rebased directly onto `main` so this PR contains only the 3 conformance-fix commits.

### DIFF
--- a/Libraries/Connect/Internal/Streaming/URLSessionStream.swift
+++ b/Libraries/Connect/Internal/Streaming/URLSessionStream.swift
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import Foundation
+import os.log
 
 /// Stream implementation that wraps a `URLSession` stream.
 ///
@@ -21,6 +22,7 @@ import Foundation
 final class URLSessionStream: NSObject, @unchecked Sendable {
     private let closedByServer = Locked(false)
     private let readStream: Foundation.InputStream
+    private let requestBodyResendRequested = Locked(false)
     private let requestBodyStreamVended = Locked(false)
     private let responseCallbacks: ResponseCallbacks
     private let task: URLSessionUploadTask
@@ -42,11 +44,24 @@ final class URLSessionStream: NSObject, @unchecked Sendable {
     /// `Assertion failed: (CFReadStreamGetStatus(_stream.get()) == kCFStreamStatusNotOpen),`
     /// `function _onqueue_setupStream_block_invoke, file HTTPRequestBody.cpp`
     ///
-    /// To avoid the crash, the stream is vended only once. Subsequent requests return `nil`, which
-    /// fails the task with an error that is surfaced through the normal stream-close path instead.
+    /// To avoid the crash, the stream is vended only once. On subsequent requests the task is
+    /// canceled and `nil` is returned. Cancellation is required in addition to returning `nil`:
+    /// CFNetwork does not treat a `nil` body stream as a failure and can instead leave the task
+    /// parked indefinitely waiting for a body stream, never delivering any delegate callback.
+    /// Canceling guarantees `urlSession(_:task:didCompleteWithError:)` is delivered so the failure
+    /// is surfaced through the normal stream-close path.
     var requestBodyStream: Foundation.InputStream? {
         return self.requestBodyStreamVended.perform { vended in
             guard !vended else {
+                os_log(
+                    .error,
+                    """
+                    URLSession requested a resend of a streamed request body which \
+                    cannot be replayed - canceling the request
+                    """
+                )
+                self.requestBodyResendRequested.value = true
+                self.task.cancel()
                 return nil
             }
             vended = true
@@ -138,11 +153,18 @@ final class URLSessionStream: NSObject, @unchecked Sendable {
 
         self.closedByServer.value = true
         if let error = error {
-            let code = Code.fromURLSessionCode((error as NSError).code)
+            var code = Code.fromURLSessionCode((error as NSError).code)
+            var message = error.localizedDescription
+            if code == .canceled && self.requestBodyResendRequested.value {
+                // The cancelation came from `requestBodyStream` refusing a resend, not from the
+                // caller. Surface it as a retriable connection failure instead.
+                code = .unavailable
+                message = "connection was reset and the streamed request body cannot be replayed"
+            }
             self.responseCallbacks.receiveClose(
                 code,
                 [:],
-                ConnectError(code: code, message: error.localizedDescription)
+                ConnectError(code: code, message: message)
             )
         } else {
             self.responseCallbacks.receiveClose(.ok, [:], nil)
--- a/Libraries/Connect/Internal/Streaming/URLSessionStream.swift
+++ b/Libraries/Connect/Internal/Streaming/URLSessionStream.swift
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 import Foundation
-import os.log
 
 /// Stream implementation that wraps a `URLSession` stream.
 ///
@@ -53,13 +52,6 @@ final class URLSessionStream: NSObject, @unchecked Sendable {
     var requestBodyStream: Foundation.InputStream? {
         return self.requestBodyStreamVended.perform { vended in
             guard !vended else {
-                os_log(
-                    .error,
-                    """
-                    URLSession requested a resend of a streamed request body which \
-                    cannot be replayed - canceling the request
-                    """
-                )
                 self.requestBodyResendRequested.value = true
                 self.task.cancel()
                 return nil
--- a/Libraries/ConnectNIO/Public/NIOHTTPClient.swift
+++ b/Libraries/ConnectNIO/Public/NIOHTTPClient.swift
@@ -262,13 +262,9 @@ open class NIOHTTPClient: Connect.HTTPClientInterface, @unchecked Sendable {
                 channel.close(mode: .all, promise: nil)
             }
         }
-        // `deinit` can run on one of `loopGroup`'s own event loop threads: an `EventLoopFuture`
-        // callback that captures `[weak self]` (such as the connect callback in
-        // `connectChannelAndMultiplexerIfNeeded()`) briefly upgrades it to a strong reference,
-        // and if the client's other owners released it in the meantime, dropping that reference
-        // deallocates the client on the event loop. `syncShutdownGracefully()` traps with a
-        // precondition failure when called from an event loop, so use the asynchronous shutdown,
-        // which is safe to invoke from any thread.
+        // An event loop callback can release the client's last reference, running `deinit` on that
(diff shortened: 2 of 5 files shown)


Libraries/Connect/Internal/Streaming/URLSessionStream.swift
```diff
@@ -42,11 +44,24 @@ final class URLSessionStream: NSObject, @unchecked Sendable {
...
+    /// To avoid the crash, the stream is vended only once. On subsequent requests the task is
+    /// canceled and `nil` is returned. Cancellation is required in addition to returning `nil`:
+    /// CFNetwork does not treat a `nil` body stream as a failure and can instead leave the task
+    /// parked indefinitely waiting for a body stream, never delivering any delegate callback.
+    /// Canceling guarantees `urlSession(_:task:didCompleteWithError:)` is delivered so the failure
+    /// is surfaced through the normal stream-close path.
     var requestBodyStream: Foundation.InputStream? {
         return self.requestBodyStreamVended.perform { vended in
             guard !vended else {
+                os_log(
```
jhump: Is this log really appropriate? I don't know how this gets surfaced in the app or on a device, but I'd worry that it could be noisy if this is something that can simply happen with an unreliable network.

Tests/ConformanceClient/Sources/main.swift
```diff
@@ -44,6 +49,13 @@ private func main() async throws {
         let request = try Connectrpc_Conformance_V1_ClientCompatRequest(
             serializedBytes: nextRequestData
         )
+        logToStderr(
```
jhump: You might consider only emitting this when a "verbose" flag is provided. That is how some of the other impls work (including the reference impls in Go and the one in kotlin). That way, in a normal run, the conformance output is tidy/quiet. But the log verbosity can be turned on, for troubleshooting, by providing a separate command-line flag to the conformance client program. (Like the `--some-flag-for-client-program` example in the docs [here](https://github.com/connectrpc/conformance/blob/main/docs/configuring_and_running_tests.md#running-tests).)

Libraries/ConnectNIO/Public/NIOHTTPClient.swift
```diff
@@ -262,7 +262,14 @@ open class NIOHTTPClient: Connect.HTTPClientInterface, @unchecked Sendable {
...
         }
-        try? self.loopGroup.syncShutdownGracefully()
+        // `deinit` can run on one of `loopGroup`'s own event loop threads: an `EventLoopFuture`
+        // callback that captures `[weak self]` (such as the connect callback in
+        // `connectChannelAndMultiplexerIfNeeded()`) briefly upgrades it to a strong reference,
+        // and if the client's other owners released it in the meantime, dropping that reference
+        // deallocates the client on the event loop. `syncShutdownGracefully()` traps with a
+        // precondition failure when called from an event loop, so use the asynchronous shutdown,
+        // which is safe to invoke from any thread.
+        self.loopGroup.shutdownGracefully { _ in }
```
rebello95, line 268: I don't really understand this change - is it supposed to be altering behavior?

Tests/ConformanceClient/Sources/main.swift
```diff
@@ -19,6 +19,11 @@ import SwiftProtobuf
 
 private let prefixLength = MemoryLayout<UInt32>.size // 4
 
+private func logToStderr(_ message: String) {
+    let timestamp = ISO8601DateFormatter().string(from: Date())
```
rebello95: We should cache this date formatter; they're expensive to instantiate

Libraries/Connect/Internal/Streaming/URLSessionStream.swift
eseay: It would only show up in the console logger, but you're right that it's probably not super valuable if it's not going to be easily read one way or the other. I'll nix it.

Tests/ConformanceClient/Sources/main.swift
eseay: Great suggestion.

Tests/ConformanceClient/Sources/main.swift
eseay: Great point - this actually pointed out an even better solution which was to use the modern `FormatStyle` API when available. 

What I've done instead is diverge the functionality based on available system version, and on modern systems, use the current `FormatStyle` API, which handles caching. For the sake of the compiler and the available deployment targets, I'll preserve the `ISO8601DateFormatter` path, but I've added a comment explaining that it's not something that's actually hit in practice (meaning that the added, albeit light, complexity of caching one isn't necessary).

Libraries/ConnectNIO/Public/NIOHTTPClient.swift
eseay: @rebello95 this is the main crux of the change that the agent was able to find by digging inside of NIO. I've updated the comment to be a bit more clear, but concise. This is how the agent explained it to me:

> syncShutdownGracefully() blocks the calling thread until the event loop has stopped.
> 
> The problem occurred when:
> 
> - A NIO future callback temporarily retained the client.
> - That callback released the client’s final reference.
> - deinit therefore ran on the event-loop thread.
> - deinit called syncShutdownGracefully().
>
> The event loop would then be waiting synchronously for itself to stop—a deadlock. SwiftNIO detects this situation and intentionally triggers a precondition failure, crashing the process. try? cannot catch that failure.
> 
> The asynchronous version initiates shutdown and returns, allowing the event-loop callback to finish so the loop can stop normally.

rebello95: I see; it's switching to the async version, got it

Tests/ConformanceClient/Sources/main.swift
```diff
@@ -19,6 +19,20 @@ import SwiftProtobuf
...
 
+private func logToStderr(_ message: String) {
+    let date = Date()
+    let timestamp: String
+    if #available(macOS 12, iOS 15, watchOS 8, tvOS 15, *) {
+        timestamp = date.formatted(.iso8601)
+    } else {
+        // This is a valid path on supported older systems, but the conformance client is not run
+        // on them in practice. If that changes, cache this formatter because it is expensive to
+        // create.
```
rebello95: Let's just make this a `private let` globally alongside `prefixLength`, I don't think we need to add a todo like this

Makefile
```diff
@@ -78,8 +78,8 @@ $(BIN)/license-headers: Makefile
 testconformance: ## Run all conformance tests
 	swift build -c release --product ConnectConformanceClient
 	mv ./.build/release/ConnectConformanceClient $(BIN)
-	PATH="$(abspath $(BIN)):$(PATH)" connectconformance --trace --conf ./Tests/ConformanceClient/InvocationConfigs/urlsession.yaml --mode client $(BIN)/ConnectConformanceClient httpclient=urlsession
-	PATH="$(abspath $(BIN)):$(PATH)" connectconformance --trace --conf ./Tests/ConformanceClient/InvocationConfigs/nio.yaml --mode client $(BIN)/ConnectConformanceClient httpclient=nio
+	PATH="$(abspath $(BIN)):$(PATH)" connectconformance --trace --conf ./Tests/ConformanceClient/InvocationConfigs/urlsession.yaml --mode client $(BIN)/ConnectConformanceClient httpclient=urlsession verbose=false
+	PATH="$(abspath $(BIN)):$(PATH)" connectconformance --trace --conf ./Tests/ConformanceClient/InvocationConfigs/nio.yaml --mode client $(BIN)/ConnectConformanceClient httpclient=nio verbose=false
```
rebello95: Do we need this? Presumably false is the default

rebello95 (approved):
lgtm after comments

rebello95 (approved):

PR #411, opened 2026-07-21
Migrate NIOHTTPClient off deprecated HTTP2StreamMultiplexer

### LINKED ISSUE #408: Migrate NIOHTTPClient off deprecated HTTP2StreamMultiplexer to clear Sendable warnings
## Summary

After bumping `swift-nio` (2.92.2 → 2.101.3), `swift-nio-http2` (1.39.0 → 1.44.0), and `swift-nio-ssl` (2.36.0 → 2.37.2), the build succeeds but the newer NIO surfaces **4 `Sendable` warnings** in [`Libraries/ConnectNIO/Public/NIOHTTPClient.swift`](../blob/main/Libraries/ConnectNIO/Public/NIOHTTPClient.swift). These were intentionally left out of the dependency-bump PR because clearing them requires a behavior-sensitive HTTP/2 refactor that should be its own, separately-tested change.

## Warnings

| Location (approx.) | Warning |
|---|---|
| `createBootstrap`, ~L85 | `conformance of 'NIOSSLHandler' to 'Sendable' is unavailable` — from `channel.pipeline.addHandler(sslHandler)` |
| `createBootstrap`, ~L91 | `conformance of 'HTTP2StreamMultiplexer' to 'Sendable' is unavailable` — from the future `.flatMap { … }.map { (_: HTTP2StreamMultiplexer) in }` chain |
| `startMultiplexChannel`, ~L238 (×2) | `type 'any ChannelHandler' does not conform to 'Sendable'` **and** `capture of 'handlers' with non-Sendable type '[any ChannelHandler]' in a '@Sendable' closure` — from `createStreamChannel(promise:) { channel in channel.pipeline.addHandlers(handlers) }` |

## Root cause

Newer swift-nio / -http2 / -ssl declare `@available(*, unavailable) extension <Type>: Sendable {}` for `NIOSSLHandler`, `HTTP2StreamMultiplexer`, and `ChannelHandler`, and the stream-initializer typealias `NIOChannelInitializer` is `@Sendable`. The file's existing `@preconcurrency import NIOCore/NIOSSL` does **not** suppress *explicitly-unavailable* Sendable conformances or the `#SendableClosureCaptures` diagnostic, so these 4 surface.

## Why there's no clean partial fix

- Warning ~L85 alone **is** cleanly fixable: replace the future-based `pipeline.addHandler(sslHandler)` with `pipeline.syncOperations.addHandler(…)` inside `channel.eventLoop.makeCompletedFuture { … }` (runs on the event loop).
- Warnings ~L91 and ~L238 **cannot** be cleanly fixed while keeping the legacy `HTTP2StreamMultiplexer`. The only **public synchronous** `configureHTTP2Pipeline` (in `NIOHTTP2/HTTP2PipelineHelpers.swift`, on `ChannelPipeline.SynchronousOperations`) returns the **inline** `NIOHTTP2Handler.StreamMultiplexer`, not the legacy type; the sync variant returning the legacy multiplexer is `internal`. The only public ways to obtain the legacy multiplexer are the future-based overloads that trigger the L91 warning.

**Conclusion:** properly clearing L91 + L238 means migrating `NIOHTTPClient` off the deprecated `HTTP2StreamMultiplexer` to the inline `NIOHTTP2Handler.StreamMultiplexer`.

## Proposed migration scope

- [ ] Change `State.connected(channel:multiplexer:)` and the `pendingRequests` closure type from `HTTP2StreamMultiplexer` to `NIOHTTP2Handler.StreamMultiplexer`.
- [ ] Update `sendOrQueueRequest`, `flushOrFailPendingRequests`, and `startMultiplexChannel` signatures accordingly.
- [ ] In `createBootstrap`'s `channelInitializer`, use `channel.eventLoop.makeCompletedFuture { … }` with `syncOperations.addHandler(sslHandler)` + `syncOperations.configureHTTP2Pipeline(mode: .client) { … }`, and **store the returned inline multiplexer** instead of later retrieving it via `syncOperations.handler(type: HTTP2StreamMultiplexer.self)` in `connectChannelAndMultiplexerIfNeeded`.
- [ ] Update `startMultiplexChannel` to use the inline multiplexer's `createStreamChannel` and construct/add the per-request handlers **on the event loop** (`syncOperations.addHandlers`) so no non-Sendable handlers are captured in the `@Sendable` initializer (may need `NIOLoopBound`, or building the handlers inside the initializer).

## Validation

- [ ] `swift build` is warning-free.
- [ ] `make testunit` passes.
- [ ] `make installconformancerunner && make testconformance` passes (this is behavior-sensitive HTTP/2 plumbing).
- [ ] Example apps (SPM + CocoaPods) still build.

## Notes

- Should ship as its own PR, separate from the dependency bump.
- Unrelated to the known URLSession single-test conformance flake tracked elsewhere — don't conflate.

eseay:
Fixes #408.

## Summary

Building `ConnectNIO` against recent `swift-nio-http2` releases produces a series of Sendable and deprecation warnings originating in `NIOHTTPClient`. The root cause is that the client was still built on `NIOHTTP2.HTTP2StreamMultiplexer`, which NIO has deprecated in favor of the inline multiplexer, `NIOHTTP2Handler.StreamMultiplexer`. This PR migrates `NIOHTTPClient` to the supported API, which eliminates all of the warnings without changing the client's public API or behavior.

## Changes

**Adopt `NIOHTTP2Handler.StreamMultiplexer`.** All references to the deprecated `HTTP2StreamMultiplexer` (the `pendingRequests` queue, the `State.connected` case, and the `sendOrQueueRequest`/`flushOrFailPendingRequests`/`startMultiplexChannel` signatures) now use `NIOHTTP2Handler.StreamMultiplexer`. After the connection is established, the multiplexer is retrieved via `pipeline.syncOperations.handler(type: NIOHTTP2Handler.self).syncMultiplexer()`, which is the supported pattern for the inline multiplexer.

**Configure the pipeline synchronously in the channel initializer.** The `channelInitializer` previously chained `EventLoopFuture`s (`pipeline.addHandler(...)` / `channel.configureHTTP2Pipeline(...)`), which both used the deprecated multiplexer under the hood and required passing non-`Sendable` values into `@Sendable` closures. It now wraps the setup in `eventLoop.makeCompletedFuture { ... }` and uses `pipeline.syncOperations` to add the SSL handler and configure the HTTP/2 pipeline. Since the initializer runs on the channel's event loop, the synchronous API is safe here, avoids the future-chaining entirely, and lets thrown errors (e.g. from `NIOSSLContext` creation) propagate through the completed future instead of the previous manual `do`/`catch` + `channel.close()` fallback.

**Create stream channels the same way.** `startMultiplexChannel` now builds its per-stream handlers inside `eventLoop.makeCompletedFuture { ... }` using `syncOperations.addHandlers(...)`, matching the connection-level setup. The `connectHandler` parameter is now `any ChannelInboundHandler & Sendable`, which the inline multiplexer's `@Sendable` stream initializer requires; both concrete handler types passed here (`ConnectUnaryChannelHandler` and `ConnectStreamChannelHandler`) already satisfy it.

**Remove `@preconcurrency` imports.** With the deprecated API gone, the `@preconcurrency` annotations on `import NIOCore` and `import NIOSSL` are no longer needed and have been removed, so any future concurrency issues in this file will surface as warnings rather than being suppressed.

## Why this approach

The deprecation message on `HTTP2StreamMultiplexer` directs users to `NIOHTTP2Handler.StreamMultiplexer`, and the inline multiplexer is where NIO's Sendable-correct APIs live. Migrating to it fixes the warnings at the source rather than suppressing them, keeps `ConnectNIO` on NIO's supported path, and requires no changes for consumers of `NIOHTTPClient`.

## Validation

- `swift build` completes with no warnings in `ConnectNIO`.
- Unit tests pass (`make testunit`).
- Conformance tests pass for both the URLSession and NIO clients (`make testconformance`).
- The example apps (Swift PM and CocoaPods) build successfully.

### DIFF
--- a/Libraries/ConnectNIO/Public/NIOHTTPClient.swift
+++ b/Libraries/ConnectNIO/Public/NIOHTTPClient.swift
@@ -15,11 +15,11 @@
 import Connect
 import Foundation
 import NIOConcurrencyHelpers
-@preconcurrency import NIOCore
+import NIOCore
 import NIOHTTP1
 import NIOHTTP2
 import NIOPosix
-@preconcurrency import NIOSSL
+import NIOSSL
 import os.log
 
 /// HTTP client powered by Swift NIO which supports trailers (unlike URLSession).
@@ -32,13 +32,13 @@ open class NIOHTTPClient: Connect.HTTPClientInterface, @unchecked Sendable {
     private let timeout: TimeInterval?
     private let useSSL: Bool
 
-    private var pendingRequests = [(NIOHTTP2.HTTP2StreamMultiplexer?) -> Void]()
+    private var pendingRequests = [(NIOHTTP2.NIOHTTP2Handler.StreamMultiplexer?) -> Void]()
     private var state = State.disconnected
 
     private enum State {
         case disconnected
         case connecting
-        case connected(channel: NIOCore.Channel, multiplexer: NIOHTTP2.HTTP2StreamMultiplexer)
+        case connected(channel: NIOCore.Channel, multiplexer: NIOHTTP2.NIOHTTP2Handler.StreamMultiplexer)
     }
 
     /// Designated initializer for the client.
@@ -74,28 +74,21 @@ open class NIOHTTPClient: Connect.HTTPClientInterface, @unchecked Sendable {
             .channelOption(ChannelOptions.socket(IPPROTO_TCP, TCP_NODELAY), value: 1)
             .channelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
             .channelInitializer { channel in
-                do {
-                    let channelPipeline: EventLoopFuture<Void>
+                channel.eventLoop.makeCompletedFuture {
                     if let tlsConfiguration = tlsConfiguration {
                         let sslContext = try NIOSSL.NIOSSLContext(configuration: tlsConfiguration)
                         let sslHandler = try NIOSSL.NIOSSLClientHandler(
                             context: sslContext, serverHostname: host
                         )
-                        channelPipeline = channel.pipeline
-                            .addHandler(sslHandler)
-                    } else {
-                        channelPipeline = channel.pipeline
-                            .addHandlers([])
+                        try channel.pipeline.syncOperations.addHandler(sslHandler)
                     }
-
-                    return channelPipeline.flatMap {
-                        return channel.configureHTTP2Pipeline(mode: .client) { channel in
-                            return channel.eventLoop.makeSucceededVoidFuture()
-                        }
+                    _ = try channel.pipeline.syncOperations.configureHTTP2Pipeline(
+                        mode: .client,
+                        connectionConfiguration: .init(),
+                        streamConfiguration: .init()
+                    ) { channel in
+                        channel.eventLoop.makeSucceededVoidFuture()
                     }
-                    .map { (_: NIOHTTP2.HTTP2StreamMultiplexer) in }
-                } catch {
-                    return channel.close(mode: .all)
                 }
             }
     }
@@ -173,7 +166,9 @@ open class NIOHTTPClient: Connect.HTTPClientInterface, @unchecked Sendable {
 
     // MARK: - Private
 
-    private func sendOrQueueRequest(send: @escaping (NIOHTTP2.HTTP2StreamMultiplexer?) -> Void) {
+    private func sendOrQueueRequest(
+        send: @escaping (NIOHTTP2.NIOHTTP2Handler.StreamMultiplexer?) -> Void
+    ) {
         self.lock.withLock {
             switch self.state {
             case .connected(_, let multiplexer):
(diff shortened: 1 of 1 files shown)


Libraries/ConnectNIO/Public/NIOHTTPClient.swift
```diff
@@ -74,28 +76,21 @@ open class NIOHTTPClient: Connect.HTTPClientInterface, @unchecked Sendable {
...
-                        channelPipeline = channel.pipeline
-                            .addHandlers([])
+                        try channel.pipeline.syncOperations.addHandler(sslHandler)
                     }
-
-                    return channelPipeline.flatMap {
-                        return channel.configureHTTP2Pipeline(mode: .client) { channel in
-                            return channel.eventLoop.makeSucceededVoidFuture()
-                        }
+                    _ = try channel.pipeline.syncOperations.configureHTTP2Pipeline(
```
rebello95, line 87: Double-checking that we don't need this result anywhere?

Libraries/ConnectNIO/Public/NIOHTTPClient.swift
```diff
@@ -15,11 +15,11 @@
 import Connect
 import Foundation
 import NIOConcurrencyHelpers
-@preconcurrency import NIOCore
```
rebello95, line 18: nice

rebello95 (approved):

Libraries/ConnectNIO/Public/NIOHTTPClient.swift
eseay: Correct. The returned value is basically just a handle/view that gives us access to the handler that gets created by this function. The pipeline itself is what holds the actual reference. We don't need to do anything with the returned value. This is logically analogous to what was being done with the `.map` call where we were just ignoring the multiplexer previously.

PR #412, opened 2026-07-22
Fix conformance URLSession timeout burst from Cache.db thrash

eseay:
## Problem
`run-conformance-tests` fails intermittently on macOS CI (seen on #406 and on `main` post-#407): one nameless `timed out waiting for result from client`, then a burst of ~97 unrelated `FAILED: … failed to get result from client` lines, always during the URLSession suite. The burst is not ~97 real failures — the serial conformance client wedged on one case and every queued case behind it timed out.

## Diagnosis
- Reproduced locally by soaking the URLSession invocation config; runs wedged intermittently mid-suite (e.g. on `server-stream/invalid-argument`), yet every wedging case passed reliably in isolation — pointing to cross-case accumulation, not a per-case bug.
- At the wedge point the client log showed CFNetwork `stepSQLStatement … Cache.db` / `SQLITE_READONLY` errors. The conformance client constructed a new `URLSessionHTTPClient()` (with `URLSessionConfiguration.default`) for every case; hundreds of default sessions in one process thrash the shared on-disk URLCache (`~/Library/Caches/<bundle>/Cache.db`). Under that corruption, CFNetwork can leave a streamed upload task parked forever without delivering any delegate callback.
- #407's fixes (async NIO shutdown, cancel-on-resend) addressed real bugs but not this one, which is why the timeout burst persisted after it merged.

This PR contains two independent changes: the root-cause fix for the CI flakiness, and a lock-scope hardening in `URLSessionHTTPClient` for a potential deadlock noticed while tracing the wedge.

## Change 1: shared ephemeral session (the flakiness fix)
- **`Tests/ConformanceClient/Sources/ConformanceInvoker.swift`** — use one process-wide `URLSessionHTTPClient` built from an ephemeral configuration (`urlCache = nil`, `.reloadIgnoringLocalCacheData`) instead of a fresh `.default` client per case. No on-disk cache, no cross-case accumulation.
- **`Libraries/Connect/Internal/Streaming/URLSessionStream.swift`** — comment-only: soften the claim that canceling on a body-stream resend *guarantees* `didCompleteWithError`; under URLCache corruption CFNetwork may delay or omit it.
- **`Tests/UnitTests/.../URLSessionStreamTests.swift`** — regression test covering the resend path end to end: second `needNewBodyStream` returns `nil` and cancels, the cancelation is remapped to `.unavailable` (not `.canceled`), and `receiveClose` fires exactly once. #407 only asserted the second vend returned `nil`.

## Change 2: lock-scope hardening (found during analysis)
- **`Libraries/Connect/Public/Implementation/Clients/URLSessionHTTPClient.swift`** — while auditing paths that could stall the delegate queue, we noticed `needNewBodyStream` held the client lock across `stream.requestBodyStream`. Since #407 that call is no longer a passive read — on a resend it calls `task.cancel()` into CFNetwork — so the lock (also taken by `didCompleteWithError` stream removal and the unary/metrics paths) was held across CFNetwork internals and the completion handler, a setup for lock-order inversion. Not the observed hang (the serial delegate queue enqueues rather than re-enters), but worth closing: the lock now covers only the dictionary lookup, and vending/canceling happens outside it. Behavior-preserving on the normal path.

## Validation
- Pre-fix: URLSession soak reproduced the timeout burst (run 3).
- Post-fix: URLSession soak 40/40 clean (no timeouts, no Cache.db errors), NIO soak 10/10, full `make testconformance` 3/3, unit suite 70/70.
- `swift build -Xswiftc -warnings-as-errors` and SwiftLint clean on touched files.

### DIFF
--- a/Libraries/Connect/Internal/Streaming/URLSessionStream.swift
+++ b/Libraries/Connect/Internal/Streaming/URLSessionStream.swift
@@ -50,7 +50,7 @@ final class URLSessionStream: NSObject, @unchecked Sendable {
     /// Canceling is the best available signal to request
     /// `urlSession(_:task:didCompleteWithError:)` so the failure can be surfaced through the
     /// normal stream-close path (CFNetwork still may delay or omit completion under process-wide
-    /// URLCache corruption; conformance avoids that by using a shared ephemeral session).
+    /// URLCache corruption.
     var requestBodyStream: Foundation.InputStream? {
         return self.requestBodyStreamVended.perform { vended in
             guard !vended else {
--- a/Libraries/Connect/Internal/Streaming/URLSessionStream.swift
+++ b/Libraries/Connect/Internal/Streaming/URLSessionStream.swift
@@ -47,8 +47,10 @@ final class URLSessionStream: NSObject, @unchecked Sendable {
     /// canceled and `nil` is returned. Cancellation is required in addition to returning `nil`:
     /// CFNetwork does not treat a `nil` body stream as a failure and can instead leave the task
     /// parked indefinitely waiting for a body stream, never delivering any delegate callback.
-    /// Canceling guarantees `urlSession(_:task:didCompleteWithError:)` is delivered so the failure
-    /// is surfaced through the normal stream-close path.
+    /// Canceling is the best available signal to request
+    /// `urlSession(_:task:didCompleteWithError:)` so the failure can be surfaced through the
+    /// normal stream-close path (CFNetwork still may delay or omit completion under process-wide
+    /// URLCache corruption; conformance avoids that by using a shared ephemeral session).
     var requestBodyStream: Foundation.InputStream? {
         return self.requestBodyStreamVended.perform { vended in
             guard !vended else {
--- a/Tests/ConformanceClient/Sources/ConformanceInvoker.swift
+++ b/Tests/ConformanceClient/Sources/ConformanceInvoker.swift
@@ -69,7 +69,12 @@ final class ConformanceInvoker: Sendable {
                 port: Int(request.port)
             )
         case .urlSession:
-            return URLSessionHTTPClient()
+            // One shared ephemeral session for the whole process. Creating a fresh
+            // `URLSessionConfiguration.default` client per case opens the process-wide
+            // on-disk URLCache (`…/Caches/<bundle>/Cache.db`) hundreds of times; under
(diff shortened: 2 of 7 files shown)


Libraries/Connect/Internal/Streaming/URLSessionStream.swift
```diff
@@ -47,8 +47,10 @@ final class URLSessionStream: NSObject, @unchecked Sendable {
     /// canceled and `nil` is returned. Cancellation is required in addition to returning `nil`:
     /// CFNetwork does not treat a `nil` body stream as a failure and can instead leave the task
     /// parked indefinitely waiting for a body stream, never delivering any delegate callback.
-    /// Canceling guarantees `urlSession(_:task:didCompleteWithError:)` is delivered so the failure
-    /// is surfaced through the normal stream-close path.
+    /// Canceling is the best available signal to request
+    /// `urlSession(_:task:didCompleteWithError:)` so the failure can be surfaced through the
+    /// normal stream-close path (CFNetwork still may delay or omit completion under process-wide
+    /// URLCache corruption; conformance avoids that by using a shared ephemeral session).
```
rebello95: talking about conformance here is a bit odd, as this is production code

Tests/ConformanceClient/Sources/ConformanceInvoker.swift
```diff
@@ -69,7 +69,7 @@ final class ConformanceInvoker: Sendable {
                 port: Int(request.port)
             )
```
rebello95, line 70: Should we do the same for `NIOHTTPClient` (sharing across runs)?

rebello95 (commented):
generally LGTM, one question

Libraries/Connect/Internal/Streaming/URLSessionStream.swift
eseay: That's a good call out. Much more appropriate for a PR comment or something.

Tests/ConformanceClient/Sources/ConformanceInvoker.swift
eseay: We theoretically could change some code to do that, but it's not necessary since the crux of the issue here was that we were blowing up the `URLCache.shared` that's associated with `URLSessionConfiguration.default`.

Different `NIOHTTPClient` instances don't share a cache the same way, so in trying to keep the scope as limited as possible, I just left it alone.

rebello95 (approved):

PR #414, opened 2026-07-28
Remove CocoaPods support

eseay:
CocoaPods Trunk [stops accepting new podspecs on December 2, 2026](https://blog.cocoapods.org/CocoaPods-Specs-Repo/), so the publishing path is going away regardless. This removes CocoaPods support and makes Swift Package Manager the only supported distribution channel.

- Delete `Connect-Swift.podspec`, `Connect-Swift-Mocks.podspec`, and the `ElizaCocoaPodsApp` example
- Drop the `build-eliza-cocoapods-example` CI job and the `publish-podspecs` release job
- Unwrap the four `#if !COCOAPODS` guards in `MenuView.swift` — a no-op, since `COCOAPODS` was only ever defined by the generated Pods xcconfigs
- Replace the three cocoapods.org README badges and add a note linking the announcement
- Simplify the release and dependency-update steps in `CONTRIBUTING.md`, which no longer touch podspecs

The already-published 1.2.3 pods remain on Trunk and keep working; they simply stop receiving updates. Nothing is pushed to Trunk as part of this change.

Since this drops a distribution channel, it should ship in the next major release.

**For a maintainer:** the `COCOAPODS_TRUNK_TOKEN` repo secret is now unused and can be revoked.


Examples/ElizaSwiftPackageApp/ElizaSwiftPackageApp/AppSources/MenuView.swift
```diff
@@ -13,11 +13,8 @@
 // limitations under the License.
 
 import Connect
-#if !COCOAPODS
-// SwiftNIO (and gRPC) support is not available via CocoaPods since SwiftNIO does not support it.
 // This import is only necessary if using gRPC, not for Connect or gRPC-Web.
```
rebello95, line 16: I think we can kill `ElizaSharedSources` altogether and move them to the SPM example app now

README.md
```diff
@@ -2,9 +2,15 @@ Connect-Swift
...
 
 [![Build](https://github.com/connectrpc/connect-swift/actions/workflows/ci.yml/badge.svg?branch=main)](https://github.com/connectrpc/connect-swift/actions/workflows/ci.yml)
-[![Version](https://img.shields.io/cocoapods/v/Connect-Swift.svg?style=flat)](https://cocoapods.org/pods/Connect-Swift)
-[![Platform](https://img.shields.io/cocoapods/p/Connect-Swift.svg?style=flat)](https://cocoapods.org/pods/Connect-Swift)
-[![License](https://img.shields.io/cocoapods/l/Connect-Swift.svg?style=flat)](https://cocoapods.org/pods/Connect-Swift)
+[![Version](https://img.shields.io/github/v/tag/connectrpc/connect-swift?label=version&style=flat)](https://github.com/connectrpc/connect-swift/releases/latest)
+[![Platform](https://img.shields.io/badge/platform-iOS%20%7C%20macOS%20%7C%20tvOS%20%7C%20watchOS-lightgrey?style=flat)](./Package.swift)
+[![License](https://img.shields.io/github/license/connectrpc/connect-swift?style=flat)](./LICENSE)
+
+> [!NOTE]
```
rebello95: I'd suggest moving this to the bottom or under `Integrating with Swift Package Manager`; probably don't want a deprecation warning near the top of the readme

rebello95 (approved):
thanks! lgtm after 2 comments

Examples/ElizaSwiftPackageApp/ElizaSwiftPackageApp/AppSources/MenuView.swift
eseay: Good point. I've gone ahead and just rebuilt the project altogether.

PR #423, opened 2026-07-28
Fix non-atomic duplicate-callback guard in UnaryAsyncWrapper

### LINKED ISSUE #422: `UnaryAsyncWrapper`'s duplicate-callback guard is not atomic and can still crash
## Summary

The guard in [`UnaryAsyncWrapper`](https://github.com/connectrpc/connect-swift/blob/7dd82ff790cc5a5bcceef18e05fb29cfa26cca4a/Libraries/Connect/Internal/Unary/UnaryAsyncWrapper.swift#L56-L75) that exists to prevent a `CheckedContinuation` from being resumed twice does not actually prevent it. Under genuinely concurrent duplicate callbacks, both callers pass the check and the process traps.

**Reproduced locally**, crashing with:

```
_Concurrency/CheckedContinuation.swift:172: Fatal error: SWIFT TASK CONTINUATION MISUSE:
send() tried to resume its continuation more than once
```

## The defect

```swift
let hasResumed = Locked(false)
self.cancelable = self.sendUnary { response in
    guard !hasResumed.value else {            // (1) read under lock
        os_log(.fault, "...")
        return
    }
    continuation.resume(returning: response)  // (2) resume — outside any lock
    hasResumed.perform(action: { $0 = true }) // (3) write under lock
}
```

(1) and (3) are two separate lock acquisitions with the resume *between* them. Two threads entering at the same instant both read `false` at (1), both proceed to (2), and the continuation is resumed twice. The flag is set after the damage is done.

### Reproducer

Drive the completion closure from two threads released at the same deadline:

```swift
let start = DispatchTime.now() + .milliseconds(20)
for _ in 0..<2 {
    DispatchQueue.global().async {
        while DispatchTime.now() < start {}
        completion(ResponseMessage(result: .success(.init())))
    }
}
```

This traps on the first or second attempt on an M-series machine.

## Fix

Collapse the check and the set into one atomic compare-and-set, and resume only after winning it. This repo already contains the correct pattern — `TimeoutHTTPClient` in [`TimeoutTests.swift`](https://github.com/connectrpc/connect-swift/blob/7dd82ff790cc5a5bcceef18e05fb29cfa26cca4a/Tests/UnitTests/ConnectLibraryTests/ConnectTests/TimeoutTests.swift#L72-L80) does exactly this:

```swift
let hasResumed = Locked(false)
self.cancelable = self.sendUnary { response in
    let shouldResume = hasResumed.perform { hasResumed -> Bool in
        if hasResumed { return false }
        hasResumed = true
        return true
    }
    guard shouldResume else {
        os_log(.fault, "`sendUnary` delivered a duplicate callback; ignoring.")
        return
    }
    continuation.resume(returning: response)
}
```

Note that `resume` stays *outside* the `perform` block. `Locked` is backed by `os_unfair_lock`, and resuming a continuation under it risks priority inversion.

That is the whole fix — a few lines, no structural change.

## Do NOT remove the actor

Worth recording, because it looks like dead weight and is not:

`UnaryAsyncWrapper` is an `actor` whose only isolated state is one `Cancelable?`, and `onCancel` hops through an unstructured `Task(priority: .high) { await self.cancelable?.cancel() }`. That reads like ceremony worth deleting. It is not — **the actor's serialization is load-bearing.**

`self.cancelable` is assigned only *after* `sendUnary` returns. If cancelation could be processed in that window, it would find `nil` and be silently dropped, leaving the request running after the caller canceled. The actor prevents this: the `Task` spawned in `onCancel` must acquire actor isolation, so it queues behind the job that performs the assignment and always observes a non-`nil` `cancelable`.

I tested this specifically — canceling while `sendUnary` was still dispatching, 100 iterations: **0 lost cancelations**. The current code is correct here.

**If you ever do replace the actor** (to drop the unstructured `Task` hop, which is the only real cost), the invariant that must be preserved is: *a cancelation arriving before the `Cancelable` exists must be recorded and honored once it does.* A `Locked` state machine over `pending / active(Cancelable) / canceled / resumed` handles it — `onCancel` marks `.canceled` when still `.pending`, and the dispatch path cancels immediately if it finds `.canceled` on completion. That is a strictly larger change than the fix above and should not be bundled with it.

## Also worth a small edit: the stale comment

The comment above the guard explains the duplicate callback as a live swift-nio defect:

> In some circumstances where a request timeout and a server error occur at nearly the same moment, the underlying `swift-nio` system will trigger this callback twice.

That root cause has since been addressed — `ConnectUnaryChannelHandler` now carries `isClosed` guards at several points. Keep the guard (a double resume is a hard crash and a correct guard costs one lock acquisition), but reword the comment as defense-in-depth rather than a workaround for a bug that is still live.

*Caveat:* this assumes the `isClosed` guards are present, which they are on `eddie/ios-13-min-version`. Verify before adopting that wording on an older branch.

## Sequencing

- **Fully independent — land it first.** One internal file, no public API, and no `swift-tools-version` bump (unlike #418, #419, and #421). Nothing else in flight touches this file.
- **Survives #420 (retiring the callback API surface).** That issue removes the *consumer-facing* callback API but keeps `HTTPClientInterface` callback-based, so `UnaryAsyncWrapper` remains the bridge between the async surface and the transport. Do not fold the two together.

## Verification

- [ ] Add a regression test firing the completion closure from two threads released at a common deadline; assert no crash and exactly one resume. Run it 50+ times — it is a race, so a single green run proves nothing.
- [ ] Add a test that a normal single response resumes exactly once.
- [ ] Add a test that cancelation during dispatch still reaches the underlying `Cancelable`, to lock in the behavior the actor currently provides so a future refactor cannot regress it silently.
- [ ] `make testunit` and `make testconformance` (both `urlsession` and `nio`) pass.
- [ ] `swiftlint lint --strict` clean.

## Acceptance criteria

- The resume guard is a single atomic compare-and-set; the check and the set cannot be separated.
- `continuation.resume` is never called while the lock is held.
- The concurrent-duplicate-callback reproducer passes repeatedly.
- The actor is retained, with a comment explaining why, *or* replaced by a state machine that provably preserves the pending-cancel invariant — with a test proving it.

eseay:
Fixes #422.

## Problem

`UnaryAsyncWrapper`'s guard against a duplicate transport callback resuming its `CheckedContinuation` twice does not prevent it. The check and the set are two separate lock acquisitions with the resume between them:

```swift
guard !hasResumed.value else { … }            // (1) read under lock
continuation.resume(returning: response)      // (2) resume — outside any lock
hasResumed.perform(action: { $0 = true })     // (3) write under lock
```

Two callbacks arriving concurrently both read `false` at (1), both proceed to (2), and the process traps with `SWIFT TASK CONTINUATION MISUSE: send() tried to resume its continuation more than once`. The flag is set after the damage is done.

## Changes

- **`UnaryAsyncWrapper.swift`** — claiming the right to resume is now a single `Locked.perform` compare-and-set, the same pattern `TimeoutHTTPClient` already uses in `TimeoutTests`. `continuation.resume` deliberately stays *outside* the lock: `Locked` is backed by `os_unfair_lock`, and resuming a continuation while holding it risks priority inversion.
- **`UnaryAsyncWrapper.swift`** — document why the actor is load-bearing. `cancelable` is assigned only after `sendUnary` returns, and actor isolation is what makes the `Task` spawned in `onCancel` queue behind that assignment instead of observing `nil` and silently dropping the cancelation. It reads like deletable ceremony, so the invariant is now recorded in the source.
- **`UnaryAsyncWrapperTests.swift`** — new suite covering the race, the normal single-callback path, and cancelation arriving mid-dispatch.

No public API change, no `swift-tools-version` bump — one internal file plus tests.

## On the stale-comment note in #422

#422 suggested softening the `swift-nio` comment to defense-in-depth on the grounds that the root cause had been fixed. It has not been, so the comment is kept as a description of a live defect. `userInboundEventTriggered` in both `ConnectUnaryChannelHandler` and `ConnectStreamChannelHandler` calls `closeConnection()` — which early-returns when already closed — and then invokes the response callback unconditionally, so a response followed by an `IdleStateEvent` delivers two callbacks. `NIOHTTPClient` installs `IdleStateHandler(allTimeout:)` whenever a timeout is configured, so the path is reachable. On the streaming side it is masked by `ProtocolClient`'s `hasCompleted`, which is why only unary traps.

That is a separate change in a separate module and will be filed on its own. This fix is required regardless: a double resume is a hard crash, so the guard has to hold even once the transport stops duplicating.

## Verification

- [x] The regression test reproduces the original trap — reverting the compare-and-set to the previous check/resume/set ordering crashes with the exact `SWIFT TASK CONTINUATION MISUSE` message from #422.
- [x] New suite green across 15 consecutive runs (1,500 duplicate-callback and 750 cancelation iterations). It is a race, so the repetition is the test; a single green run proves nothing.
- [x] `swift test` — 73 tests, 16 suites, all pass.
- [x] `make testconformance` — 966 passed / 0 failed (urlsession), 1681 passed / 0 failed (nio).
- [x] `swiftlint lint --strict` — 0 violations in 99 files.

### DIFF
--- a/Tests/UnitTests/ConnectLibraryTests/ConnectTests/UnaryAsyncWrapperTests.swift
+++ b/Tests/UnitTests/ConnectLibraryTests/ConnectTests/UnaryAsyncWrapperTests.swift
@@ -1,0 +1,116 @@
+// Copyright 2022-2025 The Connect Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+@testable import Connect
+import Foundation
+import Testing
+
+private typealias TestMessage = Connectrpc_Conformance_V1_UnaryResponse
+
+struct UnaryAsyncWrapperTests {
+    /// `UnaryAsyncWrapper` may only resume its continuation once; a second resume is a hard trap
+    /// (`SWIFT TASK CONTINUATION MISUSE`), so a regression here surfaces as a crashed test process
+    /// rather than a failed expectation.
+    ///
+    /// This is a race, and any single iteration proves nothing - the repetition is the test.
+    @Test
+    func concurrentDuplicateCallbacksResumeContinuationOnce() async {
+        for _ in 0..<100 {
+            let invocations = Locked(0)
+            let wrapper = UnaryAsyncWrapper<TestMessage> { completion in
+                // Release both callbacks at a common deadline so they land as close to
+                // simultaneously as the hardware allows.
+                let releaseTime = DispatchTime.now() + .milliseconds(5)
+                for _ in 0..<2 {
+                    DispatchQueue.global().async {
+                        while DispatchTime.now() < releaseTime {}
+                        invocations.perform { $0 += 1 }
+                        completion(ResponseMessage(result: .success(TestMessage())))
+                    }
+                }
+                return Cancelable {}
+            }
+
+            let response = await wrapper.send()
+            #expect(response.code == .ok)
+            #expect(response.message != nil)
+
+            // `send()` returns as soon as the first callback wins, so wait for the loser to
+            // finish. Asserting on this ensures the test fails loudly if the duplicate callback
+            // was never actually delivered, rather than passing vacuously.
+            while invocations.value < 2 {
+                await Task.yield()
+            }
+            #expect(invocations.value == 2)
+        }
+    }
+
+    @Test
+    func singleCallbackResumesWithItsResponse() async {
+        let invocations = Locked(0)
+        let expectedMessage = TestMessage.with { $0.payload.data = Data(repeating: 42, count: 4) }
+        let wrapper = UnaryAsyncWrapper<TestMessage> { completion in
+            invocations.perform { $0 += 1 }
+            completion(ResponseMessage(code: .ok, result: .success(expectedMessage)))
+            return Cancelable {}
+        }
+
+        let response = await wrapper.send()
+        #expect(invocations.value == 1)
+        #expect(response.code == .ok)
+        #expect(response.error == nil)
+        #expect(response.message == expectedMessage)
+    }
+
+    /// `UnaryAsyncWrapper.cancelable` is assigned only after `sendUnary` returns, and the actor's
+    /// serialization is what keeps a cancelation arriving in that window from being dropped. This
+    /// locks in that behavior so a future refactor cannot regress it silently.
+    @Test
+    func cancelationDuringDispatchReachesCancelable() async {
+        for _ in 0..<50 {
+            let didCancel = Locked(false)
+            let dispatchStarted = Locked(false)
+            let wrapper = UnaryAsyncWrapper<TestMessage> { completion in
+                // Signal that `sendUnary` is executing, then stall so the cancelation below
+                // lands before `cancelable` has been assigned.
+                dispatchStarted.value = true
+                Thread.sleep(forTimeInterval: 0.005)
+
+                // Fallback response so that a dropped cancelation fails an expectation instead
+                // of hanging the test forever.
+                DispatchQueue.global().asyncAfter(deadline: .now() + .milliseconds(250)) {
+                    completion(ResponseMessage(code: .ok, result: .success(TestMessage())))
+                }
+
+                return Cancelable {
+                    didCancel.value = true
+                    completion(ResponseMessage(
+                        code: .canceled, result: .failure(ConnectError.canceled())
+                    ))
+                }
+            }
+
+            let task = Task { await wrapper.send() }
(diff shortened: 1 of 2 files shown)


Tests/UnitTests/ConnectLibraryTests/ConnectTests/UnaryAsyncWrapperTests.swift
```diff
@@ -0,0 +1,116 @@
...
+import Testing
+
+private typealias TestMessage = Connectrpc_Conformance_V1_UnaryResponse
+
+struct UnaryAsyncWrapperTests {
+    /// `UnaryAsyncWrapper` may only resume its continuation once; a second resume is a hard trap
+    /// (`SWIFT TASK CONTINUATION MISUSE`), so a regression here surfaces as a crashed test process
+    /// rather than a failed expectation.
+    ///
+    /// This is a race, and any single iteration proves nothing - the repetition is the test.
```
rebello95: Not sure how I feel about these kinds of tests (iterating 50-100x attempting to catch a potential race condition). I'd be ok with leaving them out and just applying the fix in the other file

rebello95 (approved):
one comment about tests

Tests/UnitTests/ConnectLibraryTests/ConnectTests/UnaryAsyncWrapperTests.swift
eseay: Reasonable feedback. I kept the one that doesn't exercise that same looping behavior, but I got rid of the other two.

PR #425, opened 2026-07-28
Fix EventLoop-shutdown errors from late work in the NIO client

eseay:
`make testconformance` floods stderr with:

```
ERROR: Cannot schedule tasks on an EventLoop that has already shut down.
This will be upgraded to a forced crash in future SwiftNIO versions.
```

27 occurrences per run on `main` (e2f8e69). SwiftNIO's `fatalError` path already exists behind `SWIFTNIO_STRICT`, so a NIO upgrade turns this into a hard crash. It is residue from [#407](https://github.com/connectrpc/connect-swift/pull/407), which converted the `deinit` trap into this print without fixing the lifetime bug underneath.

## Cause

`NIOHTTPClient` owns a private event loop group and shuts it down in `deinit` without waiting for anything. Two kinds of work then land on a dead loop:

1. **Late cancelation from our own code** (26 of the 27). Handlers hold no reference to the client, so a `cancel()` arriving after teardown reaches `eventLoop.submit(...)` on a dead loop. The conformance client builds a client per test request, hence a flood.
2. **NIO's own scheduling during a connect** (the last 1). `connect(host:port:)` resolves DNS on an offload queue and hops back with `loop.execute` from inside NIO, so no gate of ours can reach it. Reproduced deterministically: 300 clients torn down mid-connect produce exactly 300 occurrences.

## Fix

A new internal `EventLoopGroupOwner` owns the group and gates scheduling onto it.

- Handlers hold it **weakly**, so a late call finds nothing and is dropped. Strong references would pin an event loop thread per request forever, since handlers are not reliably released today (`ProtocolClient`'s stream ⟷ handler cycle, `TimeoutTimer.cancel()`).
- The lock spans the enqueue. NIO accepts tasks while a loop is `.open` or `.closing`, so an enqueue that happens-before `shutdownGracefully()` is guaranteed to drain; checking a flag first reopens the race.
- `execute`, not `submit(_:)`, whose unfulfilled promise trips `EventLoopFuture.deinit`'s leaked-promise trap in debug builds.
- For cause 2, the connect is registered as outstanding work and the group's shutdown is deferred until it settles — bounded by `ClientBootstrap`'s 10s `connectTimeout`.

Nothing changes on the drop path: NIO already swallowed these actions, so they never ran. They are now dropped cleanly instead of printing and leaking a promise, which is why all 2,647 conformance cases still pass unchanged.

No public API change. The internal `init(host:port:timeout:eventLoopGroup:)` seam is for tests; an injected group is never shut down by the client, so two clients can share one.

## Verification

- **Conformance: 27 → 1 (gate) → 0 (connect lifetime)**, green on CI with both suites completing. `SWIFTNIO_STRICT=1 make testconformance` is clean.
- **The 300-client teardown repro: 300 → 0.**
- **The tests fail before the fix.** Reverting `runOnEventLoop` trips `leaking promise created at EventLoop.swift:1025` and kills the test process; dropping the outstanding-work check fails `deferShutDownWhileWorkIsOutstanding`. Two anti-regression cases pass in every state, guarding against a gate that is too aggressive.
- New `ConnectNIOTests` target, 14 tests — on a real `MultiThreadedEventLoopGroup`, since `EmbeddedEventLoop` always reports `inEventLoop == true` and never takes the path this bug lives on. `swift test` 84/84, clean over 100 consecutive runs.
- Thread count plateaus at 9 during a conformance run rather than climbing; a per-client leak across 1681 cases would show hundreds.

Two notes for reviewers. `run-conformance-tests` was already passing — `make testconformance` builds `-c release`, where this is only a print — so this removes the flood and the future-crash exposure rather than turning a red job green. And `hasInitiatedGroupShutDown` exists purely so the deferral is testable: `shutdownGracefully` is asynchronous and a closing loop still accepts tasks, so probing the loop instead would race.

### DIFF
--- a/Libraries/ConnectNIO/Internal/ConnectUnaryChannelHandler.swift
+++ b/Libraries/ConnectNIO/Internal/ConnectUnaryChannelHandler.swift
@@ -24,6 +24,9 @@ final class ConnectUnaryChannelHandler: NIOCore.ChannelInboundHandler, @unchecke
     private let request: Connect.HTTPRequest<Data?>
     private let onMetrics: (Connect.HTTPMetrics) -> Void
     private let onResponse: (Connect.HTTPResponse) -> Void
+    /// Weak by design: the handler outlives the client which created it. See
+    /// `EventLoopGroupOwner`.
+    private weak var loopGroupOwner: EventLoopGroupOwner?
 
     private var context: NIOCore.ChannelHandlerContext?
     private var isClosed = false
@@ -35,11 +38,13 @@ final class ConnectUnaryChannelHandler: NIOCore.ChannelInboundHandler, @unchecke
     init(
         request: Connect.HTTPRequest<Data?>,
         eventLoop: NIOCore.EventLoop,
+        loopGroupOwner: EventLoopGroupOwner,
         onMetrics: @escaping (Connect.HTTPMetrics) -> Void,
         onResponse: @escaping (Connect.HTTPResponse) -> Void
     ) {
         self.request = request
         self.eventLoop = eventLoop
+        self.loopGroupOwner = loopGroupOwner
         self.onMetrics = onMetrics
         self.onResponse = onResponse
     }
@@ -67,7 +72,9 @@ final class ConnectUnaryChannelHandler: NIOCore.ChannelInboundHandler, @unchecke
         if self.eventLoop.inEventLoop {
             action()
         } else {
-            self.eventLoop.submit(action).cascade(to: nil)
+            // The owner drops the action if the client - and therefore the event loop group - is
+            // already gone, rather than scheduling onto a loop which has shut down.
+            self.loopGroupOwner?.execute(on: self.eventLoop, action)
         }
     }
 
--- a/Libraries/ConnectNIO/Internal/ConnectUnaryChannelHandler.swift
+++ b/Libraries/ConnectNIO/Internal/ConnectUnaryChannelHandler.swift
@@ -24,8 +24,7 @@ final class ConnectUnaryChannelHandler: NIOCore.ChannelInboundHandler, @unchecke
     private let request: Connect.HTTPRequest<Data?>
     private let onMetrics: (Connect.HTTPMetrics) -> Void
     private let onResponse: (Connect.HTTPResponse) -> Void
-    /// Weak by design: the handler outlives the client which created it. See
-    /// `EventLoopGroupOwner`.
+    /// Weak: the handler can outlive the client. See `EventLoopGroupOwner`.
     private weak var loopGroupOwner: EventLoopGroupOwner?
 
     private var context: NIOCore.ChannelHandlerContext?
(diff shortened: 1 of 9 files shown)


Libraries/ConnectNIO/Internal/ConnectUnaryChannelHandler.swift
```diff
@@ -35,11 +37,13 @@ final class ConnectUnaryChannelHandler: NIOCore.ChannelInboundHandler, @unchecke
     init(
         request: Connect.HTTPRequest<Data?>,
         eventLoop: NIOCore.EventLoop,
+        loopGroupOwner: EventLoopGroupOwner,
```
rebello95: Can you help me understand the reasoning for having these classes store both the `NIOCore.EventLoop` and a `EventLoopGroupOwner`, versus having our own abstraction (`EventLoopGroupOwner`, for example) own and expose the loop directly?

Libraries/ConnectNIO/Internal/EventLoopGroupOwner.swift
```diff
@@ -0,0 +1,158 @@
...
+import NIOConcurrencyHelpers
+import NIOCore
+import NIOPosix
+
+/// Owns an `EventLoopGroup`'s lifetime and gates the scheduling of work onto its loops.
+///
+/// Handlers hold this **weakly**, so a call arriving after the client is gone finds nothing and is
+/// dropped rather than landing on a shut down loop. Strong references are not an option: handlers
+/// are not reliably released today (the `ProtocolClient` stream/handler cycle and
+/// `TimeoutTimer.cancel()` both leak them), so they would pin an event loop thread forever.
```
rebello95: Should we fix these? (Also, this seems like a bit of an implementation detail)

Libraries/ConnectNIO/Internal/EventLoopGroupOwner.swift
```diff
@@ -0,0 +1,158 @@
...
+    /// Creates an owner of a new single-threaded group whose lifetime it fully manages.
+    convenience init() {
+        self.init(
+            group: NIOPosix.MultiThreadedEventLoopGroup(numberOfThreads: 1), isGroupOwned: true
+        )
+    }
+
+    /// - parameter group: The group onto whose loops this owner schedules work.
+    /// - parameter isGroupOwned: Whether to shut the group down. An injected group belongs to its
+    ///                           creator and is never shut down here, so it may be shared.
```
rebello95, line 52: It'd be helpful to clarify when this should be set to true/false

Libraries/ConnectNIO/Internal/EventLoopGroupOwner.swift
```diff
@@ -0,0 +1,158 @@
...
+
+            // `execute`, not `submit(_:)`, whose promise is never fulfilled when the enqueue fails
+            // and trips `EventLoopFuture.deinit`'s leaked-promise trap in debug builds.
+            eventLoop.execute { withExtendedLifetime(self) { action() } }
+            return true
+        }
+    }
+
+    /// Registers work whose completion NIO schedules itself, bypassing `execute(on:_:)` - namely a
+    /// connect, whose DNS lookup hops back onto the loop from an offload queue. Shutdown is
```
rebello95, line 88: Seems like a bit of an implementation detail?

Libraries/ConnectNIO/Internal/EventLoopGroupOwner.swift
```diff
@@ -0,0 +1,158 @@
...
+            return self.isGroupOwned && self.outstandingWorkCount == 0
+        }
+
+        if shouldShutDownGroup {
+            self.shutDownGroup()
+        }
+    }
+
+    private func shutDownGroup() {
+        // Called outside `lock`, so retaking it here is safe.
```
rebello95, line 138: Might be worth documenting this in the function name or docstring

PR #426, opened 2026-07-29
Replace TimeoutTimer's GCD implementation with structured concurrency

### LINKED ISSUE #417: Replace `TimeoutTimer`'s GCD implementation with structured concurrency
## Background

Now that the minimum deployment target is iOS 13 / macOS 10.15 / tvOS 13 / watchOS 6, Swift Concurrency is unconditionally available to every consumer of the library — that set of platform versions is exactly the concurrency back-deployment floor. Internal implementation types that previously had to be written in pre-concurrency style, because they were reachable from the iOS 12 callback surface, are now free to use `Task`.

`TimeoutTimer` is the most self-contained of these. It is internal, has a three-method surface, and is not part of any public API.

## Current state

[`Libraries/Connect/Internal/Utilities/TimeoutTimer.swift`](https://github.com/connectrpc/connect-swift/blob/7dd82ff790cc5a5bcceef18e05fb29cfa26cca4a/Libraries/Connect/Internal/Utilities/TimeoutTimer.swift):

```swift
final class TimeoutTimer: @unchecked Sendable {
    private let hasTimedOut = Locked(false)
    private var onTimeout: (() -> Void)?
    private let queue = DispatchQueue(label: "connectrpc.Timeout")
    private let timeout: TimeInterval
    private var workItem: DispatchWorkItem! // Force-unwrapped to allow capturing self in init

    init?(config: ProtocolClientConfig) {
        guard let timeout = config.timeout else { return nil }
        self.timeout = timeout
        self.workItem = DispatchWorkItem { [weak self] in
            self?.hasTimedOut.value = true
            self?.onTimeout?()
        }
    }

    func start(onTimeout: @escaping () -> Void) {
        let milliseconds = Int(self.timeout * 1_000)
        self.queue.sync { self.onTimeout = onTimeout }
        self.queue.asyncAfter(deadline: .now() + .milliseconds(milliseconds), execute: self.workItem)
    }

    func cancel() { workItem.cancel() }
}
```

Three things are worth calling out:

1. **A dedicated serial `DispatchQueue` is allocated per timer, and a timer is allocated per RPC.** One is constructed in `ProtocolClient.unary` ([ProtocolClient.swift:60](https://github.com/connectrpc/connect-swift/blob/7dd82ff790cc5a5bcceef18e05fb29cfa26cca4a/Libraries/Connect/Public/Implementation/Clients/ProtocolClient.swift#L60)) and one in `createRequestCallbacks` ([ProtocolClient.swift:270](https://github.com/connectrpc/connect-swift/blob/7dd82ff790cc5a5bcceef18e05fb29cfa26cca4a/Libraries/Connect/Public/Implementation/Clients/ProtocolClient.swift#L270)), so every unary request and every stream pays for a fresh GCD queue whenever a timeout is configured.
2. **`workItem` is a force-unwrapped `DispatchWorkItem!`**, purely to work around capturing `self` before initialization completes.
3. **`onTimeout` is mutable state** synchronized through `queue.sync`, which is why the type needs `@unchecked Sendable`.

## Proposal

Replace the work item and queue with a `Task` and `Task.sleep(nanoseconds:)`. Cancellation semantics carry over directly: `Task.sleep` throws `CancellationError` when the task is canceled, which is equivalent to `DispatchWorkItem.cancel()` preventing a not-yet-started item from running. In both designs, a cancel that arrives after the callback has begun executing does not stop it.

```swift
final class TimeoutTimer: Sendable {
    private let hasTimedOut = Locked(false)
    private let task = Locked<Task<Void, Never>?>(nil)
    private let timeout: TimeInterval

    var timedOut: Bool {
        return self.hasTimedOut.value
    }

    init?(config: ProtocolClientConfig) {
        guard let timeout = config.timeout else {
            return nil
        }
        self.timeout = timeout
    }

    deinit {
        self.cancel()
    }

    func start(onTimeout: @escaping @Sendable () -> Void) {
        let nanoseconds = UInt64(self.timeout * 1_000_000_000)
        let hasTimedOut = self.hasTimedOut
        self.task.value = Task {
            do {
                try await Task.sleep(nanoseconds: nanoseconds)
            } catch {
                return // Canceled before firing.
            }
            hasTimedOut.value = true
            onTimeout()
        }
    }

    func cancel() {
        self.task.value?.cancel()
    }
}
```

This removes the `DispatchQueue`, the `DispatchWorkItem!`, the `var onTimeout`, and the `@unchecked` on the `Sendable` conformance — all stored properties become `let`, and `Task` is `Sendable`.

### Implementation notes

- **Capture `hasTimedOut`, not `self`.** The closure must capture the `Locked` directly. Capturing `self` strongly inside the `Task` would keep the timer alive until the sleep completes, defeating `deinit { self.cancel() }` — which is what currently stops an orphaned timer from firing.
- **`onTimeout` must become `@Sendable`.** Both call sites already satisfy this: [ProtocolClient.swift:153](https://github.com/connectrpc/connect-swift/blob/7dd82ff790cc5a5bcceef18e05fb29cfa26cca4a/Libraries/Connect/Public/Implementation/Clients/ProtocolClient.swift#L153) captures `cancelation` (a `Locked`) and [ProtocolClient.swift:396](https://github.com/connectrpc/connect-swift/blob/7dd82ff790cc5a5bcceef18e05fb29cfa26cca4a/Libraries/Connect/Public/Implementation/Clients/ProtocolClient.swift#L396) captures `pendingRequestCallbacks` (`@unchecked Sendable`). No call-site changes should be needed beyond the annotation.
- **`Task.sleep(for:)` is not an option.** It requires iOS 16 (verified against the current SDK); `Task.sleep(nanoseconds:)` is the correct API at an iOS 13 floor.
- **Consider the executor.** The current implementation fires on a dedicated queue at default QoS. An unstructured `Task {}` created from a synchronous context runs on the cooperative pool with unspecified priority. `start()` is invoked from interceptor callbacks that may be running on the URLSession delegate queue (`.userInitiated`) or a NIO event loop, so if timeout latency under load turns out to matter, `Task.detached(priority: .userInitiated)` is the fallback. Start with plain `Task {}` and measure only if the timing tests get flaky.

## Risks

`Task.sleep` has coarser scheduling tolerance than `DispatchQueue.asyncAfter`. For RPC deadlines this is not meaningful, but [`TimeoutTests.swift`](https://github.com/connectrpc/connect-swift/blob/7dd82ff790cc5a5bcceef18e05fb29cfa26cca4a/Tests/UnitTests/ConnectLibraryTests/ConnectTests/TimeoutTests.swift) configures `timeout: 0.01` (10ms), which is tight enough that the added slack is worth watching. Both existing tests assert only on the resulting `Code`/message and not on elapsed time, so they should be unaffected, but they are the ones to run repeatedly.

## Verification

- [ ] `swift build` and `make testunit` pass.
- [ ] `TimeoutTests.unaryRequestTimesOut` and `TimeoutTests.streamRequestTimesOut` pass, run 20+ times to check for added flakiness at the 10ms timeout.
- [ ] `make testconformance` passes for both the `urlsession` and `nio` invocation configs — the conformance suite exercises deadline propagation more thoroughly than the unit tests.
- [ ] `swiftlint lint --strict` is clean.

## Acceptance criteria

- `TimeoutTimer` conforms to `Sendable` without `@unchecked`.
- No `DispatchQueue`, `DispatchWorkItem`, or force-unwrapped storage remains in the type.
- `grep -rn "DispatchQueue\|DispatchWorkItem" Libraries/` returns only the `OperationQueue` in `URLSessionHTTPClient` (which is required by the `URLSession` delegate API and is out of scope here).

## Notes

Non-breaking and self-contained: `TimeoutTimer` is `internal` and its only consumer is `ProtocolClient`. This does not need to wait for a major release.

eseay:
Fixes #417.

`TimeoutTimer` was the last GCD consumer in `Libraries/`. Now that the deployment floor is iOS 13 / macOS 10.15 / tvOS 13 / watchOS 6 — exactly the Swift Concurrency back-deployment floor — it can use `Task` and `Task.sleep(nanoseconds:)` instead.

This removes the per-RPC `DispatchQueue`, the force-unwrapped `DispatchWorkItem!`, and the mutable `var onTimeout`. Every stored property is now a `let`, so the type conforms to `Sendable` without `@unchecked`. No call site changes were required.

Two deviations from the issue's proposal, both deliberate:

- **Cancelation is kept sticky** via a small `Locked<State>`. The old `DispatchWorkItem` was built in `init`, so a `cancel()` arriving before `start()` permanently disarmed the timer. A plain `Locked<Task?>` would make that cancel a no-op and let the later `start()` arm a live timer — and that ordering is reachable: in `createRequestCallbacks`, `setCallbacks` synchronously drains a queued `cancel()` through `receiveClose` before `start()` is reached.
- **The timeout is clamped** with `max(0, ...)`, because `ProtocolClientConfig.timeout` is public and unvalidated and `UInt64(negativeDouble)` traps where the previous `Int` conversion did not.

`onTimeout()` is still invoked outside any lock held by the type, since cancelation can re-enter `cancel()` or `deinit` — the deadlock fixed in #389.

Also adds `TimeoutTimerTests`, which the type previously lacked, covering firing, cancel-after-start, cancel-before-start, deallocation, and the requirement that `hasTimedOut` is set before the callback runs. Each case was checked against a deliberately mutated implementation to confirm it actually fails when the behavior regresses.

## Verification

- `swift build --build-tests -Xswiftc -warnings-as-errors`
- `swift test` — 75 tests pass
- Timeout suites soaked 20x to check for added flakiness and hangs at the 10ms timeout
- `swiftlint lint --strict` — 0 violations
- `make testconformance` — URLSession 966/966, NIO 1681/1681

One note on conformance: the first full urlsession run hit a flaky failure in `server-stream/error-with-no-responses`. It passed 20/20 in isolation and clean on five subsequent full runs, and the same flake shape reproduces against the unmodified implementation on `main`, so it is not introduced by this change.

### DIFF
--- a/Libraries/Connect/Internal/Utilities/TimeoutTimer.swift
+++ b/Libraries/Connect/Internal/Utilities/TimeoutTimer.swift
@@ -14,12 +14,21 @@
 
 import Foundation
 
-final class TimeoutTimer: @unchecked Sendable {
+/// Fires a callback if a configured deadline elapses before the timer is canceled.
+///
+/// Cancelation is sticky: once `cancel()` has been called the timer can never fire, even if
+/// `start(onTimeout:)` runs afterwards. `ProtocolClient` depends on this - draining pending
+/// request callbacks can synchronously reach `cancel()` before `start()`.
+final class TimeoutTimer: Sendable {
+    private enum State {
+        case ready
+        case started(Task<Void, Never>)
+        case canceled
+    }
+
     private let hasTimedOut = Locked(false)
-    private var onTimeout: (() -> Void)?
-    private let queue = DispatchQueue(label: "connectrpc.Timeout")
+    private let state = Locked<State>(.ready)
     private let timeout: TimeInterval
-    private var workItem: DispatchWorkItem! // Force-unwrapped to allow capturing self in init
 
     var timedOut: Bool {
         return self.hasTimedOut.value
@@ -31,25 +40,55 @@ final class TimeoutTimer: @unchecked Sendable {
         }
 
         self.timeout = timeout
-        self.workItem = DispatchWorkItem { [weak self] in
-            self?.hasTimedOut.value = true
-            self?.onTimeout?()
-        }
     }
 
     deinit {
         self.cancel()
     }
 
-    func start(onTimeout: @escaping () -> Void) {
-        let milliseconds = Int(self.timeout * 1_000)
-        self.queue.sync { self.onTimeout = onTimeout }
-        self.queue.asyncAfter(
-            deadline: .now() + .milliseconds(milliseconds), execute: self.workItem
-        )
+    /// Start the timer. Has no effect if `cancel()` has already been called.
+    func start(onTimeout: @escaping @Sendable () -> Void) {
+        // Clamped: `timeout` is caller-supplied, and `UInt64(negativeDouble)` traps.
+        let nanoseconds = UInt64(max(0, self.timeout * 1_000_000_000))
+        // Capturing the box rather than `self` lets `deinit` disarm an orphaned timer.
+        let hasTimedOut = self.hasTimedOut
+        let task = Task {
+            do {
+                try await Task.sleep(nanoseconds: nanoseconds)
+            } catch {
+                return
+            }
+
+            hasTimedOut.value = true
+            // Must stay outside any lock: this re-enters `cancel()`/`deinit` via cancelation,
+            // which is the deadlock fixed in #389.
+            onTimeout()
+        }
+        let wasCanceled = self.state.perform { state -> Bool in
+            switch state {
+            case .canceled:
+                return true
+            case .ready, .started:
+                state = .started(task)
+                return false
+            }
+        }
+        if wasCanceled {
+            task.cancel()
+        }
     }
 
     func cancel() {
-        workItem.cancel()
+        let task = self.state.perform { state -> Task<Void, Never>? in
+            switch state {
+            case .started(let task):
+                state = .canceled
+                return task
+            case .ready, .canceled:
+                state = .canceled
+                return nil
+            }
+        }
+        task?.cancel()
     }
 }
--- a/Libraries/Connect/Internal/Utilities/TimeoutTimer.swift
+++ b/Libraries/Connect/Internal/Utilities/TimeoutTimer.swift
@@ -50,7 +50,7 @@ final class TimeoutTimer: Sendable {
     func start(onTimeout: @escaping @Sendable () -> Void) {
         // Clamped: `timeout` is caller-supplied, and `UInt64(negativeDouble)` traps.
         let nanoseconds = UInt64(max(0, self.timeout * 1_000_000_000))
-        // Capturing the box rather than `self` lets `deinit` disarm an orphaned timer.
+        // Avoid capturing `self` so `deinit` disarms the orphaned timer.
         let hasTimedOut = self.hasTimedOut
         let task = Task {
             do {
--- a/Libraries/Connect/Internal/Utilities/TimeoutTimer.swift
+++ b/Libraries/Connect/Internal/Utilities/TimeoutTimer.swift
@@ -16,22 +16,27 @@ import Foundation
 
 /// Fires a callback if a configured deadline elapses before the timer is canceled.
 ///
-/// Cancelation is sticky: once `cancel()` has been called the timer can never fire, even if
-/// `start(onTimeout:)` runs afterwards. `ProtocolClient` depends on this - draining pending
-/// request callbacks can synchronously reach `cancel()` before `start()`.
+/// `.timedOut` and `.canceled` are both terminal, so cancelation is sticky - once `cancel()` has
+/// been called the timer can never fire, even if `start(onTimeout:)` runs afterwards.
+/// `ProtocolClient` depends on this - draining pending request callbacks can synchronously reach
+/// `cancel()` before `start()`.
 final class TimeoutTimer: Sendable {
     private enum State {
         case ready
         case started(Task<Void, Never>)
+        case timedOut
         case canceled
     }
 
-    private let hasTimedOut = Locked(false)
     private let state = Locked<State>(.ready)
     private let timeout: TimeInterval
 
     var timedOut: Bool {
-        return self.hasTimedOut.value
+        if case .timedOut = self.state.value {
+            return true
+        }
+
+        return false
     }
 
     init?(config: ProtocolClientConfig) {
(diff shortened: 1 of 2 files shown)


Libraries/Connect/Internal/Utilities/TimeoutTimer.swift
```diff
@@ -14,12 +14,21 @@
...
+/// `start(onTimeout:)` runs afterwards. `ProtocolClient` depends on this - draining pending
+/// request callbacks can synchronously reach `cancel()` before `start()`.
+final class TimeoutTimer: Sendable {
+    private enum State {
+        case ready
+        case started(Task<Void, Never>)
+        case canceled
+    }
+
     private let hasTimedOut = Locked(false)
```
rebello95: Do we still need this separate variable or can we fold it into `State` with a new `timedOut` case? I think that might be nicer since we'd avoid a second lock and keep state centralized

Libraries/Connect/Internal/Utilities/TimeoutTimer.swift
```diff
@@ -31,25 +40,55 @@ final class TimeoutTimer: @unchecked Sendable {
...
+        let task = Task {
+            do {
+                try await Task.sleep(nanoseconds: nanoseconds)
+            } catch {
+                return
+            }
+
+            hasTimedOut.value = true
+            // Must stay outside any lock: this re-enters `cancel()`/`deinit` via cancelation,
+            // which is the deadlock fixed in #389.
```
rebello95: Let's avoid adding stateful repo-related comments like this

Libraries/Connect/Internal/Utilities/TimeoutTimer.swift
```diff
@@ -31,25 +40,55 @@ final class TimeoutTimer: @unchecked Sendable {
...
-        let milliseconds = Int(self.timeout * 1_000)
-        self.queue.sync { self.onTimeout = onTimeout }
-        self.queue.asyncAfter(
-            deadline: .now() + .milliseconds(milliseconds), execute: self.workItem
-        )
+    /// Start the timer. Has no effect if `cancel()` has already been called.
+    func start(onTimeout: @escaping @Sendable () -> Void) {
+        // Clamped: `timeout` is caller-supplied, and `UInt64(negativeDouble)` traps.
+        let nanoseconds = UInt64(max(0, self.timeout * 1_000_000_000))
+        // Capturing the box rather than `self` lets `deinit` disarm an orphaned timer.
```
rebello95:
```suggestion
        // Avoid capturing `self` so `deinit` disarms the orphaned timer.
```

rebello95 (approved):

Libraries/Connect/Internal/Utilities/TimeoutTimer.swift
eseay: You can definitely fold it into state, but it adds a little bit of code. I think that's a worthwhile tradeoff though. I like the idea of having all the state managed in one place.

rebello95 (approved):

PR #428, opened 2026-07-30
Adopt the `package` access level for PackageInternal APIs

### LINKED ISSUE #419: Adopt the `package` access level for `PackageInternal` APIs
## Background

The four files under [`Libraries/Connect/PackageInternal/`](https://github.com/connectrpc/connect-swift/tree/7dd82ff790cc5a5bcceef18e05fb29cfa26cca4a/Libraries/Connect/PackageInternal) implement a workaround for an access level that did not exist when they were written. Each symbol is declared `public`, prefixed with `_`, given a doc comment disclaiming stability, and annotated with a fake deprecation:

```swift
/// **This should not be considered part of Connect's public/stable interface, and is subject
/// to change. When the compiler supports it, this should be package-internal.**
@available(
    swift,
    deprecated: 100.0,
    message: "This is an internal-only API which will be made package-private in Swift 6."
)
public static func _packMessage(...)
```

The compiler has supported it since Swift 5.9 (SE-0386). The remaining blocker was CocoaPods, which built the library as a single module and could not express the same-package relationship — and CocoaPods support was removed in #414.

## Scope

**9 declarations across 4 files**, each carrying its own 5-line `@available` block and stability disclaimer:

| File | Symbols |
|---|---|
| `PackageInternal/Envelope.swift` | `Envelope` (the enum itself), `_prefixLength`, `_packMessage`, `_unpackMessage`, `_isCompressed`, `_messageLength` |
| `PackageInternal/Headers+GRPC.swift` | `_addingGRPCHeaders` |
| `PackageInternal/Trailers+gRPC.swift` | `_grpcStatus` |
| `PackageInternal/ConnectError+GRPC.swift` | `_parseGRPCHeaders` |

Roughly 38 call sites reference them, split between `Libraries/Connect/Internal/Interceptors/` (same module) and `Libraries/ConnectNIO/Internal/GRPCInterceptor.swift` (different module, same package).

**`ConnectNIO` is the only cross-module consumer**, and it is in the same SwiftPM package — which is precisely the case `package` is designed for.

## Prerequisite: bump `swift-tools-version` to 5.9

This is a hard blocker, not a nicety. SwiftPM only passes `-package-name` to the compiler when the manifest declares tools-version 5.9 or newer. I verified this with a two-target scratch package:

```
########## tools-version 5.6 ##########
error: the package access level used on 'coreOnly()' requires a package name;
       set it with the compiler flag -package-name
########## tools-version 5.9 ##########
Build complete!
```

As noted in #418 (`AsyncStream.makeStream`), the current `// swift-tools-version:5.6` declaration is already inaccurate — the library uses parameterized existentials and constrained extension sugar that require Swift 5.7+, and CI runs Xcode 26.6. Whichever of the two issues lands first should carry the bump.

## Proposal

1. Bump `swift-tools-version` in `Package.swift` from 5.6 to 5.9.
2. Change the 9 declarations from `public` to `package`.
3. Delete all 9 `@available(swift, deprecated: 100.0, ...)` blocks and the "should not be considered part of Connect's public/stable interface" disclaimers — the access level now states this in the type system.
4. Drop the `_` prefixes and update the ~38 call sites:
   - `Envelope._prefixLength` → `Envelope.prefixLength`
   - `Envelope._packMessage` → `Envelope.packMessage`
   - `Envelope._unpackMessage` → `Envelope.unpackMessage`
   - `Envelope._isCompressed` → `Envelope.isCompressed`
   - `Envelope._messageLength` → `Envelope.messageLength`
   - `Headers._addingGRPCHeaders` → `Headers.addingGRPCHeaders`
   - `Trailers._grpcStatus` → `Trailers.grpcStatus`
   - `ConnectError._parseGRPCHeaders` → `ConnectError.parseGRPCHeaders`
5. Consider whether `PackageInternal/` still deserves to be its own directory, or whether the files belong under `Internal/` now that the distinction is expressed in code. (`Envelope` and the gRPC header helpers cross a module boundary that the other `Internal/` files do not, so keeping the directory is defensible — worth a moment's thought either way.)

### Pre-checked details

- **No name collisions.** I grepped for each de-underscored name. The only near-miss is `HeaderConstants.grpcStatus` (a `String` constant on a different type), which does not conflict with a `grpcStatus()` method on `Trailers`.
- **Generated code does not touch these APIs.** No `.connect.swift` or `.mock.swift` output references `Envelope` or any of the gRPC helpers, so consumers' generated code is unaffected and no codegen changes are required.
- **Tests keep working unchanged.** `EnvelopeTests` and friends use `@testable import Connect`, which grants access to `package` symbols the same as `internal`. Only the `_` renames need applying.
- **`ConnectConformanceClient` uses none of them** (it uses plain `import Connect`), so nothing there needs to change.

## Breaking-change assessment

This is source-breaking for anyone outside the package who is calling these symbols despite the disclaimer and the deprecation warning. That is the intended outcome, and it belongs in the same major release as the iOS 13 floor.

Worth an entry in the 2.0 release notes: these symbols were always documented as unstable and always emitted a deprecation warning, and the migration is that there is no migration — the functionality was never meant to be reachable. Anyone genuinely depending on `Envelope`'s framing helpers should open an issue so a real public API can be considered.

## Verification

- [ ] `swift build` succeeds — in particular confirm `ConnectNIO` still compiles, since it is the only cross-module consumer and the one target that proves `-package-name` is being passed.
- [ ] `make testunit` passes.
- [ ] `make testconformance` passes for both `urlsession` and `nio`.
- [ ] `grep -rn "deprecated: 100.0" Libraries/` returns nothing.
- [ ] `grep -rn "Envelope\._\|\._addingGRPCHeaders\|\._grpcStatus\|\._parseGRPCHeaders" Libraries/ Tests/` returns nothing.
- [ ] Build the Eliza example app to confirm an external-style consumer is unaffected: `xcodebuild -scheme ElizaSwiftPackageApp build`.
- [ ] `swiftlint lint --strict` is clean.

## Acceptance criteria

- Zero `@available(swift, deprecated: 100.0, ...)` annotations remain in the repository.
- Zero `_`-prefixed declarations remain in `Libraries/Connect/PackageInternal/`.
- `Connect`'s public API surface no longer exposes `Envelope`, `_addingGRPCHeaders`, `_grpcStatus`, or `_parseGRPCHeaders`.
- `ConnectNIO` builds against the `package`-scoped symbols.

## Notes

Blocked on nothing but the tools-version bump. Pairs naturally with the other 2.0 breaking changes.

eseay:
Closes #419.

Replaces the pre-5.9 `public` + underscore + fake-deprecation workaround in `PackageInternal/` with the real `package` access level, now that CocoaPods (the blocker) is gone. Bumps `swift-tools-version` to 5.9, drops underscores from the 9 renamed symbols and their ~54 call sites, and renames `Headers+GRPC.swift` to `Headers+gRPC.swift` to match `Trailers+gRPC.swift`.

Source-breaking for anyone calling these despite the disclaimer — intended, ships with the other 2.0 breaking changes (#416, #414).

### DIFF
--- a/Libraries/Connect/Internal/Interceptors/ConnectInterceptor.swift
+++ b/Libraries/Connect/Internal/Interceptors/ConnectInterceptor.swift
@@ -158,7 +158,7 @@ extension ConnectInterceptor: StreamInterceptor {
 
     @Sendable
     func handleStreamRawInput(_ input: Data, proceed: @escaping (Data) -> Void) {
-        proceed(Envelope._packMessage(input, using: self.config.requestCompression))
+        proceed(Envelope.packMessage(input, using: self.config.requestCompression))
     }
 
     @Sendable
@@ -190,7 +190,7 @@ extension ConnectInterceptor: StreamInterceptor {
                 let responseCompressionPool = self.streamResponseHeaders.value?[
                     HeaderConstants.connectStreamingContentEncoding
                 ]?.first.flatMap { self.config.responseCompressionPool(forName: $0) }
-                if responseCompressionPool == nil && Envelope._isCompressed(data) {
+                if responseCompressionPool == nil && Envelope.isCompressed(data) {
                     proceed(.complete(
                         code: .internalError, error: ConnectError(
                             code: .internalError, message: "received unexpected compressed message"
@@ -199,7 +199,7 @@ extension ConnectInterceptor: StreamInterceptor {
                     return
                 }
 
-                let (headerByte, message) = try Envelope._unpackMessage(
+                let (headerByte, message) = try Envelope.unpackMessage(
                     data, compressionPool: responseCompressionPool
                 )
                 let isEndStream = 0b00000010 & headerByte != 0
--- a/Libraries/Connect/Internal/Interceptors/GRPCWebInterceptor.swift
+++ b/Libraries/Connect/Internal/Interceptors/GRPCWebInterceptor.swift
@@ -32,12 +32,12 @@ extension GRPCWebInterceptor: UnaryInterceptor {
         proceed: @escaping (Result<HTTPRequest<Data?>, ConnectError>) -> Void
     ) {
         // gRPC-Web unary payloads are enveloped.
-        let envelopedRequestBody = Envelope._packMessage(
+        let envelopedRequestBody = Envelope.packMessage(
             request.message ?? Data(), using: self.config.requestCompression
         )
         proceed(.success(HTTPRequest(
             url: request.url,
-            headers: request.headers._addingGRPCHeaders(using: self.config, grpcWeb: true),
+            headers: request.headers.addingGRPCHeaders(using: self.config, grpcWeb: true),
             message: envelopedRequestBody,
             method: request.method,
             trailers: nil,
@@ -57,7 +57,7 @@ extension GRPCWebInterceptor: UnaryInterceptor {
         }
 
         guard let responseData = response.message, !responseData.isEmpty else {
-            let (grpcCode, connectError) = ConnectError._parseGRPCHeaders(
+            let (grpcCode, connectError) = ConnectError.parseGRPCHeaders(
                 response.headers,
                 trailers: response.trailers
             )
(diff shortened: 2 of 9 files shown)


rebello95 (approved):
awesome!

PR #431, opened 2026-07-31
Share Locked with ConnectNIO via the package access level

eseay:
## Summary

`ConnectNIO` couldn't see `Connect`'s internal `Locked` type, so it carried a private duplicate backed by `NIOLock`. That's no longer necessary: the manifest is `swift-tools-version:5.9`, and #428 already adopted `package` for `PackageInternal` types.

- Moves `Locked` into `PackageInternal/` and marks it `package`, matching the existing convention (see `Envelope`).
- Deletes the 14-line duplicate `Locked` class in `GRPCInterceptor.swift` and its now-unused `NIOConcurrencyHelpers` import.

`Lock` stays `internal` — `Locked`'s reference to it is `private`, so it doesn't need to be exposed. No call sites change.

## Test plan

- [x] `swift build && swift test` — 86/86 passing
- [x] `make testconformance` — 966/966 (URLSession) + 1681/1681 (NIO)
- [x] SwiftLint clean

### DIFF
--- a/Libraries/Connect/PackageInternal/Locked.swift
+++ b/Libraries/Connect/PackageInternal/Locked.swift
@@ -16,12 +16,12 @@ import Foundation
 
 /// Class containing an internal lock which can be used to ensure thread-safe access to an
 /// underlying value. Conforms to `Sendable`, making it accessible from `@Sendable` closures.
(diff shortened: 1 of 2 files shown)


rebello95 (approved):
awesome

PR #432, opened 2026-07-31
Add a withDeadline structured-concurrency helper

eseay:
## Summary

Adds `withDeadline`, a small structured-concurrency helper that races an operation against a deadline and cancels the loser. It's unused for now — a pure addition, laying groundwork for replacing `TimeoutTimer`'s callback-based state machine in a later change.

- `withDeadline(_:operation:)` in `Deadline.swift` races `operation` against `Task.sleep` in a `TaskGroup`, returning `nil` if the deadline wins.
- Nothing else is modified; `TimeoutTimer` and its call sites are untouched.

## Follow-up

`TimeoutTimer.swift` is a four-state machine that exists to reconcile a timer against callback-based ordering — sticky cancellation, a deinit-disarm, and a `.canceled && timedOut → .deadlineExceeded` correlation that's currently duplicated in two places in `ProtocolClient.swift`. None of that reconciliation is needed once callers can just `await withDeadline(...)`: losing the race is directly observable, so there's nothing left to correlate.

The plan is for `ProtocolClient`'s unary and streaming request paths to adopt `withDeadline` directly in a follow-up change. Once nothing references it, `TimeoutTimer.swift` and its dedicated unit tests get deleted outright. The existing end-to-end timeout test that exercises real behavior through `ProtocolClient` stays and gets adjusted in place, since it's pinning behavior rather than the state machine.

Keeping this PR scoped to just the helper — with no consumers yet — keeps it small and independently reviewable.

## Test plan

- [x] `swift build && swift test` — 92/92 passing, suite completes in ~3s
- [x] `DeadlineTests` covers: operation wins, deadline wins, no deadline, negative-timeout clamping, operation cancellation on a lost race, and a regression check that a won race returns promptly rather than blocking for the full deadline
- [x] SwiftLint clean

### DIFF
--- a/Libraries/Connect/Internal/Utilities/Deadline.swift
+++ b/Libraries/Connect/Internal/Utilities/Deadline.swift
@@ -16,18 +16,14 @@ import Foundation
 
 /// Races `operation` against a deadline, canceling it if the deadline wins.
 ///
-/// - parameter timeout: Deadline in seconds. `nil` runs `operation` with no deadline.
+/// - parameter timeout: Deadline in seconds.
 /// - parameter operation: The work to perform.
 ///
 /// - returns: The result of `operation`, or `nil` if the deadline elapsed first.
 func withDeadline<T: Sendable>(
-    _ timeout: TimeInterval?,
+    _ timeout: TimeInterval,
     operation: @escaping @Sendable () async -> T
 ) async -> T? {
-    guard let timeout else {
-        return await operation()
-    }
-
     // `UInt64(negativeDouble)` traps, so negative timeouts are clamped to zero.
     let nanoseconds = UInt64(max(0, timeout * 1_000_000_000))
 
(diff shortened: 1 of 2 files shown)


Libraries/Connect/Internal/Utilities/Deadline.swift
```diff
@@ -0,0 +1,48 @@
...
+import Foundation
+
+/// Races `operation` against a deadline, canceling it if the deadline wins.
+///
+/// - parameter timeout: Deadline in seconds. `nil` runs `operation` with no deadline.
+/// - parameter operation: The work to perform.
+///
+/// - returns: The result of `operation`, or `nil` if the deadline elapsed first.
+func withDeadline<T: Sendable>(
+    _ timeout: TimeInterval?,
```
rebello95: This should probably be non-optional, since this function provides no additive value in the nil case

rebello95 (commented):
Would be great to see a (stacked?) PR demonstrating how this will be used before merging dead code

Libraries/Connect/Internal/Utilities/Deadline.swift
eseay: Yea we can leave this open until the other stuff is done. I like that idea.

